Issue #300, opened 2023-03-23
Add `turso tutorial` command and call it from Homebrew postinstall


penberg:
Let's have a simple `turso tutorial` command that prints basic stuff like signup, login, etc, and make the Homebrew postinstall step call that instead.

CodingDoug:
The CLI tutorial can be found here, if the CLI wants to link to it:

https://docs.turso.tech/tutorials/get-started-turso-cli

StefanoSaffran:
It looks like it had already been done but it didn't work and it was reverted.
https://github.com/chiselstrike/turso-cli/commit/a895b072bf4174b9484d20ab9e9eac0f49e69b07

haaawk:
Please research what's the proper way to do it @StefanoSaffran

haaawk:
Fixed with https://github.com/chiselstrike/turso-cli/pull/385

PR #320, opened 2023-04-06
Add `turso quickstart` command

penberg:
Refs #300

### DIFF
--- a/internal/cmd/quickstart.go
+++ b/internal/cmd/quickstart.go
@@ -1,0 +1,29 @@
+package cmd
+
+import (
+	"fmt"
+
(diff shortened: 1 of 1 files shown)


glommer:
I'm ok with it if @CodingDoug likes it, but for this to be really useful it needs to be more interactive over time.

CodingDoug:
This is fine. I'm not sure I'd call this text a tutorial, though.  Maybe a guide or quickstart.

athoscouto:
One great addition to this is resetting the register flags that Glauber created [here](https://github.com/chiselstrike/turso-cli/pull/285/files).
So after calling `turso tutorial` you would also get a round of hints the first time you use each command.

PR #321, opened 2023-04-07
allow .tables and .schema without interactive shell

jmwoliver:
This allows passing in `.tables` and `.schema` without calling in to the interactive shell:

```bash
turso db shell test-db ".tables"
turso db shell test-db ".schema"
```

The checks for those statements were also moved into `query()` so they would work. Also updates some of the docs to reflect that passing in SQL as an argument to the `shell` command is possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,14 @@ You can start an interactive SQL shell similar to `sqlite3` with:
 turso db shell <database name>
 ```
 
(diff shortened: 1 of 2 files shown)


haaawk (approved):
Thanks for your contribution @jmwoliver!!!

PR #329, opened 2023-04-12
Use database token on shell and inspect commands

athoscouto:
This fixes 401 response that happens when a user logs in on a second device.


penberg:
looks good!

Issue #154, opened 2023-02-04
Any time a 401 error occurs, show suggestion to log in with `turso auth login`


CodingDoug:
Here is are a couple examples that shows less helpful error messages:

```
$ turso db show d1
Error: failed to get database listing: 401 Unauthorized
```

```
$ turso db create foo
Error: could not create database foo: Authentication required
```

CodingDoug:
We still have a lot of inconsistency with error messaging when a token is expired.  Could we survey all of the places where auth errors can occur and make them consistent, suggesting a login command at *every opportunity*?  IMO "410 Unauthorized" should never be displayed.

```
$ turso account show
Error: failed to get database listing: 401 Unauthorized

$ turso db list
Error: failed to get database listing: 401 Unauthorized

$ turso db show my-db
Error: user not logged in, please login with turso auth login

$ turso db locations
Error: unable to fetch locations: 401 Unauthorized

$ turso db shell my-db
Error: could not create turso client: user not logged in, please login with turso auth login

$ turso db replicate my-db sjc
Error: failed to get database listing: 401 Unauthorized

$ turso db inspect my-db
Error: user not logged in, please login with turso auth login

$ turso db create foo
Error: could not create database foo: Authentication required

$ turso db destroy my-db
Error: user not logged in, please login with turso auth login

$ turso db tokens create my-db -e none
Error: user not logged in, please login with turso auth login

$ turso db update my-db
Error: user not logged in, please login with turso auth login
```

penberg:
@CodingDoug There are two different cases: (1) when you have not logged in (2) when you have logged in but are unauthorized to access something. The error message for the two should be different. What we need to do is check if we have a valid access token at the beginning of every command and if not, suggest login. For everything else, we should just propagate the "unauthorized" error back to the user (but probably not show the 404).

CodingDoug:
@penberg FWIW the above set of examples shows the results while using an expired token, which is a normal, expected case.  The error messaging should be more specific for that case - we don't want to user to think they've done something wrong, we just want them to do the usual ceremony to correct it.  This leads there to be a few distinct cases:

| Case | Logged in? | Result |
| --- | --- | --- |
| No token | no | they can do nothing and need to log in first |
| Valid token, permission OK | yes | token validates and they are allowed to perform the action |
| Valid token, no permission | yes | token validates but they are not allowed to perform the action |
[...]

PR #332, opened 2023-04-17
Drop obsolete auth check in REST API client

penberg:
The CLI already validates the access token before making more REST API calls so the check is obsolete. Furthermore, the check is also incorrect because even if you're logged in, you might be unauthorized to do some things.

### DIFF
--- a/internal/turso/turso.go
+++ b/internal/turso/turso.go
@@ -64,9 +64,6 @@ func (t *Client) do(method, path string, body io.Reader) (*http.Response, error)
 	if err != nil {
 		return nil, err
 	}
-	if resp.StatusCode == http.StatusUnauthorized {
-		return nil, fmt.Errorf("%s - please login with %s", resp.Status, Emph("turso auth login"))
-	}
 	return resp, nil
 }
 


haaawk:
I'm totally fine with removing the check. Please just make sure that this does not cause regression on https://github.com/chiselstrike/turso-cli/issues/154.

One case I can see is token expiring after CLI checks the token but before the request is made.

haaawk:
@CodingDoug is a better person to comment on this.

penberg:
We check the token every time before using it by making a "dummy" REST API call with it so #154 should be fine.

haaawk:
My old homebrew installation returns:
```
╰──> turso db list
Error: failed to get database listing: 401 Unauthorized - please login with turso auth login
```

haaawk:
Another example:
```
╰──> ./turso db create tttt
Error: could not create database tttt: Authentication required
```
vs
```
╰──> turso db create tttt
Error: could not create database tttt: failed to create database: 401 Unauthorized - please login with turso auth login
```

CodingDoug:
Can we assume that every 401 is a user auth error that can be resolved by logging in again?  If so, we can always state in the error message that logging in is the next step. Also I don't see that there is need for a dummy request to validate auth, as it doesn't necessarily guarantee that future requests will succeed (the token could expire in the meantime).

penberg:
@CodingDoug @haaawk 401 means you don't have access to a resource, it doesn't mean you're not logged in. For example, if I try to delete a database that one of you owns, I will get 401. In the future, with organizations, you may have a database show up in your listing, but deleting will give you 401 if you're not the owner.

penberg:
@haaawk https://github.com/chiselstrike/turso-cli/pull/336 fixes the regression

CodingDoug:
In that case, the response could contain a more specific error code that indicates a login is required again.

PR #335, opened 2023-04-17
Allow creation of read only tokens

haaawk:


athoscouto (approved):
LGTM

Something to note though: this will enable users to generate RO tokens for all DBs, most of which won't support the read-only claim and will give full access to such tokens.

We should probably only allow issues such tokens for DBs which run on the `sqld` version that supports that feature.

haaawk:
Good point @athoscouto. Should we do the check on the API side?

athoscouto:
Ideally, those feature checks should exist on the API level.
In this case, I think we should do whatever is easier.
Since we have logic to get the `sqld` version on the CLI, it may be easier to implement it on the here.

PR #336, opened 2023-04-17
Fix access token validation

penberg:
Commit c2b284118f7579f2a07f60d260b4383069ad083a caused a regression in
the CLI because JWT validation in getAccessToken() is broken.

Before:

```
$ turso db create
Error: could not create database comic-humbug: Authentication required
```

After:

```
$ turso db create
Error: user not logged in, please login with turso auth login
```

### DIFF
--- a/internal/cmd/db.go
+++ b/internal/cmd/db.go
@@ -137,8 +137,8 @@ func getAccessToken() (string, error) {
 	}
 
 	token := settings.GetToken()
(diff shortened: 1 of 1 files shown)


haaawk (approved):

haaawk:
I still see the regression:
```
╰──> ./turso db list
Error: failed to get database listing: 401 Unauthorized
```

PR #338, opened 2023-04-19
Warn users that their platform token will not work with databases

athoscouto:
Screenshot 2023-04-19 at 12 23 33 PM

Warning is sent to `stderr`, so things like: 

Screenshot 2023-04-19 at 12 24 06 PM

Should work.

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"text/template"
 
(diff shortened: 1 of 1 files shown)


athoscouto:
From those sentences, which one do you think is the easiest to understand?

- `authenticate to Turso`
- `authenticate to Turso API`
- `authenticate to Turso platform`

I'd rather avoid mentioning the CLI, since we may expose our API directly in the short term.
Also mentioning both API and the platform makes me a bit confused.

glommer:
Agree with Athos that this token is not just for the CLI, it can be used for direct API access.
Maybe a small rewording:
"... authenticate accesses to Turso's API, including the Turso CLI."
?

CodingDoug:
If I'm coming to Turso fresh, I wouldn't really know that "Turso's API" refers to the platform API rather than the client library APIs.  They could both easily be referred to as "Turso's API" in casual conversation.

Also I don't think anyone is using the platform API directly at all, right?  So today, for consumers, it really is just the CLI in practice, until we actually document and encourage use of the platform API directly.

If we want to future-proof this, we should probably come up with a formal name for the platform API and start using it consistently.  It could be simply "the Turso platform API".  That would be more distinct from "the Turso database API" that someone might have in mind.

glommer:
@CodingDoug there are already some people using the API extra-officially, and we want to encourage more.

glommer:
I think Turso Platform API is okay.

athoscouto:
I've reworded the message a bit to mention turso platform API.
Also added info about the token kind in the cmd description, thanks for the suggestion, Doug.

PR #342, opened 2023-04-19
nudge users to book a call with us on first replica created

glommer:


athoscouto:
@glommer what about giving them both options?
Either book a call or send feedback

glommer:
now that feedback exists, I updated the PR with that.

internal/cmd/db_replicate.go
```diff
@@ -86,6 +86,13 @@ var replicateCmd = &cobra.Command{
 		fmt.Printf("   %s\n\n", dbUrl)
 		fmt.Printf("You can start an interactive SQL shell with:\n\n")
 		fmt.Printf("   turso db shell %s\n\n", dbUrl)
+
+		firstTime := config.RegisterUse("db_replicate")
+		if firstTime {
+			fmt.Printf("How is your experience going? We'd love to know!\n🗓   Book a call with us! You can do it with:\n\n\t%s\n", internal.Emph("turso account bookmeeting"))
+			fmt.Printf("🎤 Or just send us your feedback:\n\n\t%s\n", internal.Emph("turso account feedback"))
```
athoscouto, line 93:
```suggestion
			fmt.Println("How is your experience going? We'd love to know!")
			fmt.Printf("🗓   Book a call with us! You can do it with:\n\n\t%s\n", internal.Emph("turso account bookmeeting"))
			fmt.Printf("🎤   Or just send us your feedback:\n\n\t%s\n", internal.Emph("turso account feedback"))
```

PR #344, opened 2023-04-20
change help text for db replicate

glommer:
What we are doing now, is printing a replica-specific URL. This gives users the impression that they may have to connect there specifically to reach this replica.

Instead, change the text to reinforce that the main URL should work automatically. Tell them how to acquire a specific URL if they really need to.


glommer:
@CodingDoug CC

internal/cmd/db_replicate.go
```diff
@@ -81,11 +81,10 @@ var replicateCmd = &cobra.Command{
...
 		fmt.Printf("Replicated database %s to %s in %d seconds.\n\n", internal.Emph(dbName), internal.Emph(regionText), int(elapsed.Seconds()))
 
-		fmt.Printf("URL:\n\n")
-		dbUrl := getInstanceUrl(config, &database, instance)
-		fmt.Printf("   %s\n\n", dbUrl)
-		fmt.Printf("You can start an interactive SQL shell with:\n\n")
-		fmt.Printf("   turso db shell %s\n\n", dbUrl)
+		dbUrl := getDatabaseUrl(config, &database, false)
+		showCmd := fmt.Sprintf("turso db show %s", dbName)
+		fmt.Printf("Connections to %s locations physically close to %s will now be automatically routed here.\n\n", internal.Emph(dbUrl), internal.Emph(region))
```
CodingDoug, line 86: Grammar is a bit off here.  How about:

> Client connections using [dbUrl] that are physically close to [region] will now be routed to this replica.

glommer: @CodingDoug merged most of it. Are you against the word "automatic" , as in "will now be automatically routed" ?

I wanted to convey a sense of "nothing for you to do". Other than that, took your suggestions.

CodingDoug: Not necessarily against it, but I'm cautious in using it too much. Many normal behaviors of a system could be considered "automatic".  IMO, use of the word is mostly relevant when contrasting against something that would normally be done manually.  Since the routing to the nearest replica is part of the normal functioning of the product, I don't really see that adding "automatic" really helps the developer understand what's going on.

PR #353, opened 2023-04-24
fix: headless login showing error message wrongly

### LINKED ISSUE #324: Headless login is a bit broken
Let's fix the odd complaint that user is not logged in when attempting to login...

```
penberg@vonneumann ~ % turso auth login --headless
https://api.turso.io?redirect=false

Friendly reminder that there's a newer version of Turso CLI available.
You're currently using version v0.51.0 while latest available version is v0.52.0.
Please consider updating to get new features and more stable experience.

Error: user not logged in, please use turso auth login
penberg@vonneumann ~ %
```

StefanoSaffran:
This PR aims to fix the headless login showing `not logged in` error message wrongly 

closes #324

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -179,6 +179,16 @@ func auth(cmd *cobra.Command, args []string, path string) error {
 			return fmt.Errorf("error persisting username on local config: %w", err)
 		}
 		fmt.Printf("✔  Success! Logged in as %s\n", username)
+
+		firstTime := settings.RegisterUse("auth_login")
+		client, err := createTursoClient()
+		if err != nil {
+			return err
+		}
+		dbs, err := getDatabases(client)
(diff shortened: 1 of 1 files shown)


internal/cmd/auth.go
```diff
@@ -179,6 +179,16 @@ func auth(cmd *cobra.Command, args []string, path string) error {
 			return fmt.Errorf("error persisting username on local config: %w", err)
 		}
 		fmt.Printf("✔  Success! Logged in as %s\n", username)
+
+		firstTime := settings.RegisterUse("auth_login")
```
StefanoSaffran, line 183: I'm going to change it because `RegisterUse` is not being counted if it is `--headless`.

haaawk: I think it's fine to only count calls that don't have --headless.

haaawk (approved):

PR #355, opened 2023-04-25
Fix tests

haaawk:


athoscouto (approved):
Another way of doing this is using `cmd.Output()` instead of `cmd.CombinedOutput()`.
Not sure if it will break other stuff though.

This is also fine.

PR #358, opened 2023-04-25
Create a database from a seed file

glommer:
extend the creation command to add a --from-file flag that allows one to pass a sqlite file

### DIFF
--- a/internal/cmd/db_create.go
+++ b/internal/cmd/db_create.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/athoscouto/codename"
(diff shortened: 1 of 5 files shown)


internal/cmd/db_create.go
```diff
@@ -70,6 +89,14 @@ var createCmd = &cobra.Command{
 			Password: res.Password,
 		}
 
+		if dbFile != nil {
```
jmwoliver, line 92: I wonder if it makes a little more sense to have all the logic in one place after the database has been created?

```go
if dbFromFile != "" {
    dbFile, err := getDbFile(dbFromFile)
    if err != nil {
        return err
    }
    defer dbFile.Close()
    err := client.Databases.Seed(name, dbFile)
    if err != nil {
        client.Databases.Delete(name)
        return fmt.Errorf("could not create database %s: %w", name, err)
    }
}
```

jmwoliver: Maybe it is better to check that the file exists first before creating the database though, I can see why you went that route instead.

glommer: yeah, mostly we want to make sure that system-related errors don't even trigger a database creation event.

athoscouto:
Closing in favor of https://github.com/chiselstrike/turso-cli/pull/361.
PRs from forks don't work well with our CI :(
I also made some changes! 👀

PR #360, opened 2023-04-26
feat: add api tokens commands

### LINKED ISSUE #323: Add CLI support for service tokens
The API supports service tokens (we call them API tokens), but there's no `turso` CLI support for them.

StefanoSaffran:
This PR aims to add the api tokens commands to the cli. 

closes #323


internal/cmd/auth_listapitokens.go
```diff
@@ -0,0 +1,38 @@
...
+		apiTokens, err := getApiTokens(client)
+		if err != nil {
+			return err
+		}
+
+		data := [][]string{}
+		for _, apiToken := range apiTokens {
+			data = append(data, []string{apiToken.Name, fmt.Sprint(apiToken.Owner), fmt.Sprint(apiToken.PubKey)})
+		}
+		printTable([]string{"Name", "Owner", "PubKey"}, data)
```
StefanoSaffran, line 34: @athoscouto I am not sure about the API tokens use cases, Which field do we need to show in the list command? 
And the PubKey is returning empty, not sure if this is the current behavior or if I did something wrong with the unmarshal.

internal/cmd/auth.go
```diff
@@ -303,3 +310,13 @@ func logout(cmd *cobra.Command, args []string) error {
...
+
+func getApiTokens(client *turso.Client) ([]turso.ApiToken, error) {
+	return client.ApiTokens.List()
+}
+func createApiTokens(client *turso.Client, name string) (turso.CreateApiTokenResponse, error) {
+	return client.ApiTokens.Create(name)
+}
+func revokeApiTokens(client *turso.Client, name string) error {
+	return client.ApiTokens.Revoke(name)
+}
```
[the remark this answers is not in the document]
StefanoSaffran: No specific reason, I just saw this pattern with `getDatabases` and follow it.

internal/cmd/auth_listapitokens.go
StefanoSaffran: Got it.
I updated the PR with your points.

About the PubKey, I saw why it is returning empty, it is defined with `json:"-"` in the API.
Do you think we need it here? 
If so, I can open a PR changing it there.

athoscouto: We can move forward without it. If somebody asks, we add it.

internal/cmd/auth.go
```diff
@@ -81,12 +81,18 @@ var tokenCmd = &cobra.Command{
 	},
 }
 
+var apiTokensCmd = &cobra.Command{
+	Use:   "api_tokens",
+	Short: "Manage your api tokens",
```
athoscouto, line 86: You should use `API` instead of `api` in every non-error string.

athoscouto (commented):
Last but not least, I think this feature will add to the confusion users have around tokens.
I suggest you add a long description, like [this one](https://github.com/chiselstrike/turso-cli/blob/fix/271/api-token-multiple-misleading-source/internal/cmd/auth.go#L60-L62), to the commands you've just added.

You can tell users that API tokens are revocable non-expiring tokens that share the same identity as the user that creates them and can be used to implement automations using the `turso` CLI or API.

athoscouto (approved):
Have one last comment, please do it before merging.

internal/cmd/auth.go
```diff
@@ -81,12 +81,21 @@ var tokenCmd = &cobra.Command{
 	},
 }
 
+var apiTokensCmd = &cobra.Command{
+	Use:   "api_tokens",
+	Short: "Manage your API tokens",
+	Long: "" +
+		"API tokens are revocable non-expiring tokens that share the same identity \n" +
+		"that can be used to implement automations using " + internal.Emph("turso") + " CLI or API",
```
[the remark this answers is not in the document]
StefanoSaffran: Done, thanks!
I'll wait and see if @haaawk wants to add anything else.

internal/cmd/auth.go
```diff
@@ -81,12 +81,21 @@ var tokenCmd = &cobra.Command{
 	},
 }
 
+var apiTokensCmd = &cobra.Command{
+	Use:   "api_tokens",
```
haaawk, line 85: I don't think `_` in a command name is a good idea. `-` is better IMO

StefanoSaffran: Done!

internal/turso/apiTokens.go
```diff
@@ -0,0 +1,76 @@
...
+	ID     string `json:"dbId"`
+	Name   string
+	Owner  uint
+	PubKey []byte
+}
+
+type ApiTokensClient client
+
+func (a *ApiTokensClient) List() ([]ApiToken, error) {
+	res, err := a.client.Get("/v2/auth/api-tokens", nil)
```
athoscouto, line 18: Please replace `v2` with `v1`.

internal/turso/apiTokens.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+type CreateApiTokenResponse struct {
+	Name  string `json:"name"`
+	ID    string `json:"id"`
+	Token string `json:"token"`
+}
+
+func (a *ApiTokensClient) Create(name string) (CreateApiTokenResponse, error) {
+	url := fmt.Sprintf("/v2/auth/api-tokens/%s", name)
```
athoscouto, line 42: Please replace `v2` with `v1`.

internal/turso/apiTokens.go
StefanoSaffran: Done!

haaawk (approved):

PR #363, opened 2023-04-26
fix: auth token multiple misleading source

### LINKED ISSUE #271: Clarity about the source of auth token used
Right now, we have three ways of finding an auth token:

- In settings.json, from `turso auth login`
- In TURSO_API_TOKEN env var
- In `-t` CLI flag

Since all three may be present, it's potentially confusing since the CLI is not very clear about which one of these it's using when a command runs successfully.  Also, if the CLI fails because the chosen token is expired or invalid, there is no feedback about which one was problematic, leaving the user to potentially take action that doesn't resolve the issue.

Proposal:

- Let's define the order in which these sources are used. I suggest: 1. CLI flag, 2. env var, 3. settings.json
- For all (non-auth) commands that require a token, always look at all sources. If more than one token is found, immediately print a warning message listing all sources where tokens were found, and which one is being used to complete the command.

StefanoSaffran:
This PR aims to warn the user about the source of the token

closes #271

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -94,7 +94,7 @@ func isJwtTokenValid(token string) bool {
 	if len(token) == 0 {
 		return false
 	}
-	client, err := createTursoClient()
+	client, err := tursoClient(token)
 	if err != nil {
 		return false
 	}
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -181,7 +181,7 @@ func auth(cmd *cobra.Command, args []string, path string) error {
 		fmt.Printf("✔  Success! Logged in as %s\n", username)
 
 		firstTime := settings.RegisterUse("auth_login")
-		client, err := createTursoClient()
+		client, err := createTursoClientFromAccessToken(false)
 		if err != nil {
 			return err
 		}
--- a/internal/cmd/db.go
+++ b/internal/cmd/db.go
@@ -100,7 +100,7 @@ func getDatabases(client *turso.Client) ([]turso.Database, error) {
 }
 
 func completeInstanceName(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-	client, err := createTursoClient()
+	client, err := createTursoClientFromAccessToken(false)
 	if err != nil {
 		return nil, cobra.ShellCompDirectiveNoFileComp
 	}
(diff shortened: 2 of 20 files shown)


athoscouto:
The first two commits look good!

I'm not totally comfortable with the third one though.
I think the env should have priority over the settings file.
But if somebody has an env set and the token is valid, I don't think we should warn them about it.

If the token on the env is invalid, we can remember that the env is set.
If they try to login or logout, we should also warn them that this will not work as expected, since they have an env set.

I'm repeating myself since I already said those things on the issue.

But, once again, I see two downsides with those blanket warnings:
- It messes up scripts. People expect a consistent output from a command, if you add unexpected things to `stdout` it can break scripts that use our CLI. 
- It gets annoying, you're almost convincing them to not use the token env. I don't think it is good DX to have to check your settings file (God knows where it is) to remove the token before you add it to your env var.

internal/cmd/auth.go
```diff
@@ -181,7 +179,7 @@ func auth(cmd *cobra.Command, args []string, path string) error {
 		fmt.Printf("✔  Success! Logged in as %s\n", username)
 
 		firstTime := settings.RegisterUse("auth_login")
-		client, err := createTursoClient()
+		client, err := createTursoClientFromAccessToken(false)
```
haaawk, line 182: This commit does not compile. createTursoClientFromAccessToken is introduced in later commit.

internal/cmd/db.go
```diff
@@ -121,22 +121,25 @@ var dbCmd = &cobra.Command{
...
 	settings, err := settings.ReadSettings()
 	if err != nil {
 		return "", fmt.Errorf("could not read local settings")
 	}
+	settingsToken := settings.GetToken()
 
-	token := settings.GetToken()
-	if !isJwtTokenValid(token) {
+	if envToken != "" && settingsToken != "" && warnMultipleAccessTokenSources {
+		fmt.Println("Warning: found both environment and configuration token. Proceeding with environment one")
```
haaawk, line 133: I think we could improve the the error message to be more user friendly. Something like:
```
User logged in as <fetch user name from settings> but TURSO_API_TOKEN environment variable set so using it instead.
```

internal/cmd/db_passwd.go
```diff
@@ -54,7 +54,7 @@ var changePasswordCmd = &cobra.Command{
 
 		bar := prompt.Spinner("Changing password...")
 		defer bar.Stop()
-		client, err = createTursoClient()
+		client, err = createTursoClientFromAccessToken(false)
```
haaawk, line 57: I'm wondering why we can't reuse the client from line 29

haaawk:
LGTM in general. @CodingDoug do you have any opinion regarding removing the ability to specify access token using command-line flag?

CodingDoug:
> LGTM in general. @CodingDoug do you have any opinion regarding removing the ability to specify access token using command-line flag?

@haaawk I re-read my comments in the original issue and found that I couldn't come up with a strong use case for allowing the command line flag.  I suppose I still agree with my past self. :)

haaawk:
> > LGTM in general. @CodingDoug do you have any opinion regarding removing the ability to specify access token using command-line flag?
> 
> @haaawk I re-read my comments in the original issue and found that I couldn't come up with a strong use case for allowing the command line flag. I suppose I still agree with my past self. :)

Good - let's remove it then

internal/cmd/db_passwd.go
StefanoSaffran: Maybe the idea when it was created was to handle the error that wasn't handled in the first one.
https://github.com/chiselstrike/turso-cli/commit/adde529ea5e9e35ca7bbbc53f8c8f784112fe446

I tested it now removing the second client and it works.
I think it is safe to remove it, right?

haaawk: if the test passes then it should be fine.

haaawk (approved):

PR #366, opened 2023-04-27
change(shell): use libsql-shell-go library was shell

luisfvieirasilva:
`turso db shell` using libsql-shell-go lib.

Notes:
- `libsql-shell-go` is a private repo for now. So to test we need to configure it correctly (see instructions below)

----

How to add libsql-shell-go private module:
- Tell Go that libsql-shell-go is a private module, so it shouldn't search for it at https://sum.golang.org/
  - `export GOPRIVATE="github.com/libsql/libsql-shell-go"`
- Configure your GitHub to always use SSH instead of HTTPS for GitHub. You need to configure a SSH Key and add them to an agent (the key password must be saved by the agent before importing the module)
  - [Link](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent) about how configure SSH Key
  - At `~/.gitconfig` (doesn't work if configure repo's .git/config), add following lines
```
[url "ssh://git@github.com/"]
    insteadOf = https://github.com/
```
- Get mod packages
  - `go mod download`


athoscouto (changes requested):
`turso-cli` is public, it doesn't make much sense to add a private dependency to it.
Please open source the dependency before making this change.

glommer:
I am testing this and will add comments as I find them.
First one:
* this lost the spinner that @athoscouto put so much effort into cooking and was a nice DX, so we should add it back

internal/cmd/db_show.go
```diff
@@ -49,11 +49,6 @@ var showCmd = &cobra.Command{
 			return nil
 		}
 
-		if showHttpUrlFlag {
```
haaawk, line 52: Why is this removed?

internal/cmd/db_shell.go
```diff
@@ -188,224 +125,19 @@ func tokenFromDb(db *turso.Database, client *turso.Client) (string, error) {
...
-	req, err := http.NewRequest("POST", url, bytes.NewReader(body))
-	if err != nil {
-		return nil, err
-	}
-	if token != "" {
-		req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", token))
-	}
-	return http.DefaultClient.Do(req)
+func addTokenAsQueryParameter(dbUrl string, token string) string {
+	return fmt.Sprintf("%s/?jwt=%s", dbUrl, token)
```
haaawk, line 142: Do we need `/?`? I thought `/` is just needed in zsh to escape `?` but here it shouldn't be needed, no?

internal/cmd/db_shell.go
```diff
@@ -67,117 +46,75 @@ var shellCmd = &cobra.Command{
...
 		}
 
-		token, err := tokenFromDb(db, client)
+		token, err := tokenFromDb(&db, client)
 		if err != nil {
 			return err
 		}
 
 		dbUrl := nameOrUrl
-		if db != nil {
```
haaawk, line 76: Why is it now safe not to check whether db is not nil?

internal/cmd/db_shell.go
```diff
@@ -46,12 +31,6 @@ var shellCmd = &cobra.Command{
 		}
 		cmd.SilenceUsage = true
 
-		spinner := prompt.StoppedSpinner("Connecting to database")
```
haaawk, line 49: This needs to be brought back, right?

internal/cmd/db_shell.go
luisfvieirasilva: It's not zsh escape, it's just a optional final slash in the end of a path that I'm used to add, but it's unnecessary

internal/cmd/db_shell.go
luisfvieirasilva: I forced an error inside `databaseFromNameOrURL` if we can't find the database

luisfvieirasilva:
> What's blocking this @luisfvieirasilva ? Have you addressed all the comments?

I was changing libsql-client-go to be able to use `http` connections here. I'll go back to this once the [PR](https://github.com/libsql/libsql-client-go/pull/26) is merged

luisfvieirasilva (commented):
@haaawk @athoscouto @glommer PR updated

athoscouto (approved):

internal/cmd/db_shell.go
```diff
@@ -76,26 +63,34 @@ var shellCmd = &cobra.Command{
 		dbUrl := nameOrUrl
 		if db != nil {
 			dbUrl = getDatabaseHttpUrl(config, db)
+			dbUrl = addTokenAsQueryParameter(dbUrl, token)
+		}
+
+		connectionInfo := getConnectionInfo(nameOrUrl, db, config)
+
+		shellConfig := shell.ShellConfig{
+			DbPath:         dbUrl,
```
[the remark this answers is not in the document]
luisfvieirasilva: @athoscouto, new two tickets based on this comment

- [Improve DbPath name](https://github.com/libsql/libsql-shell-go/issues/87)
- [Add option to pass Bearer token as ShellConfig](https://github.com/libsql/libsql-shell-go/issues/88)

internal/cmd/db_shell.go
```diff
@@ -48,8 +33,10 @@ var shellCmd = &cobra.Command{
 		cmd.SilenceUsage = true
 
 		spinner := prompt.StoppedSpinner("Connecting to database")
+		spinnerStarted := false
 		if len(args) == 1 {
 			spinner.Start()
+			spinnerStarted = true
```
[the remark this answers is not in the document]
luisfvieirasilva: done!

PR #368, opened 2023-04-27
Organization CRSD support

athoscouto:
Let you `c`reate, list (`r`etrieve), `s`elect the current org, and `d`elete orgs.
We're still missing logic to manage users inside an org.

Some examples of command outputs:
```console
$ turso org
Manage your organizations

Usage:
  turso org [command]

Available Commands:
  create      Create a new organization
  destroy     Destroy an organization
  list        List your organizations
  select      Select an organization as the context for your commands.

Flags:
  -h, --help   help for org

Global Flags:
  -c, --config-path string   Path to the directory with config file
  -t, --token string         Access token used for authorization

Use "turso org [command] --help" for more information about a command.
```

```console
$ turso org list
NAME         SLUG                 
personal     athoscouto (current)     
test2        test2                    
```

```console
$ turso org select test2
Default organization set to test2.
All your turso commands will be executed in that organization context.
```

```console
turso org list
NAME         SLUG            
personal     athoscouto          
test2        test2 (current)     
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,6 @@ require (
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/spf13/viper v1.15.0
 	github.com/xwb1989/sqlparser v0.0.0-20180606152119-120387863bf2
-	golang.org/x/exp v0.0.0-20230310171629-522b1b587ee0
 	golang.org/x/sync v0.1.0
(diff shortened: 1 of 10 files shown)


penberg (approved):
Awesome, ship it! 👏

PR #373, opened 2023-04-28
Improve error handling for org commands

athoscouto:
Makes two improvements:
- When the user tries to manage members of orgs they aren't the owner, we tell them that it can't happen.
- When the user has an org configured to which they aren't members, we tell them that and set them back to personal org.

### DIFF
--- a/internal/turso/organizations.go
+++ b/internal/turso/organizations.go
@@ -68,19 +68,16 @@ func (c *OrganizationsClient) Delete(slug string) error {
 		return fmt.Errorf("could not find organization %s", slug)
 	}
 
-	if r.StatusCode == http.StatusBadRequest {
+	switch r.StatusCode {
+	case http.StatusOK:
+		return nil
+	case http.StatusBadRequest:
 		return fmt.Errorf("cannot delete personal organization %s", slug)
-	}
-
-	if r.StatusCode == http.StatusUnauthorized {
+	case http.StatusForbidden:
 		return fmt.Errorf("you do not have permission to delete organization %s", slug)
-	}
-
-	if r.StatusCode != http.StatusOK {
+	default:
 		return fmt.Errorf("failed to delete organization: %s", r.Status)
 	}
-
-	return nil
 }
 
 type Member struct {
--- a/internal/turso/organizations.go
+++ b/internal/turso/organizations.go
@@ -3,6 +3,9 @@ package turso
 import (
 	"fmt"
 	"net/http"
+
+	"github.com/chiselstrike/iku-turso-cli/internal"
+	"github.com/chiselstrike/iku-turso-cli/internal/settings"
 )
 
 type OrganizationsClient client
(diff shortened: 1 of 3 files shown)


internal/turso/organizations.go
```diff
@@ -100,6 +100,10 @@ func (c *OrganizationsClient) ListMembers() ([]Member, error) {
 	}
 	defer r.Body.Close()
 
+	if r.StatusCode == http.StatusForbidden {
+		return nil, fmt.Errorf("only organization owners can list membrs")
```
CodingDoug, line 104: typo: members

internal/turso/organizations.go
```diff
@@ -129,6 +133,10 @@ func (c *OrganizationsClient) AddMember(username string) error {
 	}
 	defer r.Body.Close()
 
+	if r.StatusCode == http.StatusForbidden {
+		return fmt.Errorf("only organization owners can add membrs")
```
CodingDoug, line 137: typo: members

internal/turso/organizations.go
```diff
@@ -148,6 +156,10 @@ func (c *OrganizationsClient) RemoveMember(username string) error {
 	}
 	defer r.Body.Close()
 
+	if r.StatusCode == http.StatusForbidden {
+		return fmt.Errorf("only organization owners can remove membrs")
```
CodingDoug, line 160: typo: members

internal/turso/organizations.go
```diff
@@ -161,3 +173,24 @@ func (c *OrganizationsClient) MembersURL(suffix string) (string, error) {
...
+
+func isNotMemberErr(status int, org string) bool {
+	if status == http.StatusForbidden && org != "" && unsetOrganization() == nil {
+		return true
+	}
+	return false
+}
+
+func notMemberErr(org string) error {
+	msg := fmt.Sprintf("you are not part of organization %s anymore. ", internal.Emph(org))
```
CodingDoug, line 193: The "anymore" word suggests that the user once was a member but is no longer.  I'm not so sure that's always going to be the case.  If not, just remove the word and plainly state lack of membership.

internal/turso/organizations.go
```diff
@@ -161,3 +173,24 @@ func (c *OrganizationsClient) MembersURL(suffix string) (string, error) {
...
+func isNotMemberErr(status int, org string) bool {
+	if status == http.StatusForbidden && org != "" && unsetOrganization() == nil {
+		return true
+	}
+	return false
+}
+
+func notMemberErr(org string) error {
+	msg := fmt.Sprintf("you are not part of organization %s. ", internal.Emph(org))
+	msg += fmt.Sprintf("%s configured to use your personal organization.", internal.Emph("turso"))
```
CodingDoug, line 194: I'm not sure I understand this.  This suggests to me that a change of org selection automatically happened in all cases where this function `notMemberErr` is called.

athoscouto: That is true. But after we do the change we need to communicate it to the user somehow, and that is why we return this error.

PR #375, opened 2023-04-29
Improve Windows experience

haaawk:


mateossh:
What was the purpose of this change? Were there specific bug?

haaawk:
`github.com/fatih/color` wasn't working on Windows. It was using control characters that windows shell wasn't recognising.

Issue #138, opened 2023-02-02
Hung HTTP endpoint causes CLI shell to hang, can't recover with ctl-c, can only use ctl-\ to kill


CodingDoug:
Basically I expect that I can terminate the shell (or terminate the sent SQL command, staying in the shell) with ctl-c when hung on sending a command.  Or in other words, SIGINT should get my prompt back, shouldn't need a SIGKILL.

PR #380, opened 2023-04-29
Add timeout to db inspect

haaawk:
I have an old db for which db inspect hangs for ever.
Reaction to Ctrl-C has a big delay and the overall experience
is pretty bad.

With this change db inspect times out after 30s and
prints a meaningful error.

This PR has just one commit and depends on https://github.com/chiselstrike/turso-cli/pull/379

### DIFF
--- a/internal/cmd/account_show.go
+++ b/internal/cmd/account_show.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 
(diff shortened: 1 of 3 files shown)


CodingDoug:
This sounds related to #138

haaawk:
> This sounds related to #138

Yes. Thanks @CodingDoug. There's a bit more work to make sure that all calls have sane timeouts but this is a step forward.

PR #383, opened 2023-04-30
auto update CLI

### LINKED ISSUE #266: [RFE] self updating cli
As a user I'd like to be able to update the cli client using a command like `turso update`

This is a very common feature for such tools

glommer:
Fixes #266

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -3,9 +3,7 @@ package cmd
 import (
 	"context"
 	_ "embed"
-	"encoding/json"
 	"fmt"
-	"io"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


glommer:
Note: I didn't test this on homebrew since we need to release a version with this on homebrew for the test to be accurate.
(I did test variations of this with copying the binary around, etc)

glommer:
I did =(

haaawk (approved):
LGTM but it seems that some checks are failing

glommer:
I updated it. But now it is failing with something I have no idea why =(

glommer:
it was failing on a test.. maybe it re-run by itself?

haaawk:
I rerun them. I fixed the tests today. Shouldn't be flaky any more.

PR #386, opened 2023-05-02
Fix SIGSEGV with `db replicate`

penberg:


athoscouto (approved):

internal/cmd/db_replicate.go
```diff
@@ -63,6 +63,10 @@ var replicateCmd = &cobra.Command{
 			return err
 		}
 		dbSettings := config.GetDatabaseSettings(database.ID)
+		if dbSettings == nil {
+			// Backwards compatibility with old settings files.
+			dbSettings = config.GetDatabaseSettings(database.Name)
+		}
```
haaawk, line 69: We should still check for nil in case the db is just not in the settings, no?

penberg: Sure, but the same bug exists in `getUrl()` so let's fix it in a follow up

haaawk (approved):

PR #387, opened 2023-05-02
feat: hide instance urls by default on db show command

### LINKED ISSUE #362: Don't show instance URLs by default with `turso db show`
A vast majority of the time, clients will prefer to use logical database URLs, since they should always route the client to the closes replica.  There is normally no need to use an instance URL.

In order to prevent confusion when looking at the output of `turso db show`, let's consider not showing the URLs by default, and instead require a different command or flag to show them.  We already have a command to show instance URLs alone:

```
$ turso db show $DB --instance-url $INSTANCE_NAME
```

We could also add a flag to show them in the main output:

```
$ turso db show $DB --instance-urls
```

I think I just prefer requiring the existing command to find it.

StefanoSaffran:
This PR aims to add a flag to hide the instance URLs by default on `db show` command

closes #362

### DIFF
--- a/testing/database_test.go
+++ b/testing/database_test.go
@@ -102,7 +102,7 @@ func createReplica(c *qt.C, dbName string, configPath *string, replicaName strin
 }
 
 func runSqlOnPrimaryAndReplica(c *qt.C, dbName string, configPath *string, tablePrefix string, replicaName string) {
-	output, err := turso(configPath, "db", "show", dbName)
+	output, err := turso(configPath, "db", "show", dbName, "--instance-urls")
 	c.Assert(err, qt.IsNil, qt.Commentf(output))
 	c.Assert(output, qt.Contains, "Locations:      ams, waw")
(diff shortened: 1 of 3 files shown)


haaawk (approved):

haaawk (changes requested):
Actually you need to fix tests because this change breaks them

haaawk:
You just need to add the new `instance-urls` flag to all `db show` calls in tests

testing/database_test.go
```diff
@@ -42,7 +42,7 @@ func testCreate(c *qt.C, dbName string, region *string, configPath *string, tc t
 	c.Assert(output, qt.Contains, "Created database "+dbName)
 
 	if region != nil {
-		output, err = turso(configPath, "db", "show", dbName)
+		output, err = turso(configPath, "db", "show", dbName, "--instance-urls")
```
haaawk, line 45: It seems that here we don't need the `instance-urls` flag

haaawk (approved):

PR #388, opened 2023-05-02
feat: show database size

### LINKED ISSUE #40: Show database size in `db list`
Show the database size as part of `db list`. Requires API work & integration with Fly.

StefanoSaffran:
This PR aims to show database size on `db list` and `db show` command.

closes #40 

image
image

### DIFF
--- a/internal/cmd/db_inspect.go
+++ b/internal/cmd/db_inspect.go
@@ -5,15 +5,17 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/chiselstrike/iku-turso-cli/internal/settings"
-	"github.com/dustin/go-humanize"
-	"github.com/rodaine/table"
-	"github.com/spf13/cobra"
-	"golang.org/x/sync/errgroup"
 	"io"
 	"net/http"
 	"strings"
 	"time"
+
+	"github.com/chiselstrike/iku-turso-cli/internal/settings"
+	"github.com/chiselstrike/iku-turso-cli/internal/turso"
+	"github.com/dustin/go-humanize"
+	"github.com/rodaine/table"
+	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


internal/cmd/db_inspect.go
```diff
@@ -87,38 +89,46 @@ var dbInspectCmd = &cobra.Command{
...
-		for range instances {
-			ret := <-results
-			inspectRet.Accumulate(ret)
-		}
 		inspectRet.show()
 		return nil
 	},
 }
 
+func CalculateInstancesUsedSize(instances []turso.Instance, config *settings.Settings, db turso.Database, token string) (InspectInfo, error) {
```
haaawk, line 101: Can this be used in db show as well?

internal/cmd/db_list.go
```diff
@@ -49,35 +43,10 @@ var listCmd = &cobra.Command{
...
-				})
-			}
-			if err := g.Wait(); err != nil {
-				if errors.Is(err, context.DeadlineExceeded) {
-					return fmt.Errorf("timeout while inspecting database. It's possible that this database is too old and does not support inspecting or one of the instances is not reachable")
-				}
+			inspectRet, err := CalculateInstancesUsedSize(instances, settings, database, token)
+			if err != nil {
 				return err
 			}
```
haaawk: It's better not to fail on error and just show a message that fetching the size wasn't possible.
It works the same way for fetching instance version in db show.
It's better to show what we can and inform that some data was not able to be fetched.

haaawk:
LGTM in general but I don't think we should fail the whole command when fetching size of a db fails. Let's just print a message in place of the size that says that fetching was unsuccessful and why.

internal/cmd/db_inspect.go
StefanoSaffran: Done!

internal/cmd/db_show.go
```diff
@@ -85,55 +83,31 @@ var showCmd = &cobra.Command{
...
 			if showInstanceUrlsFlag {
-				data = append(data, []string{instance.Name, instance.Type, instance.Region, version, urls[idx]})
+				data = append(data, []string{instance.Name, instance.Type, instance.Region, version, instancesInfo.urls[idx]})
 			} else {
 				data = append(data, []string{instance.Name, instance.Type, instance.Region, version})
 			}
 		}
+		var size string
+		if instancesInfo.inspectInfo == nil {
+			size = fmt.Sprintf("fetching size failed: %s", err)
```
haaawk, line 101: The value of err here seems to be something random. Probably nil if check in line 87 didn't return.

haaawk (approved):

PR #391, opened 2023-05-03
Remove password from instance creation request

athoscouto:


haaawk:
I don't understand this change. Won't it lead to replicas and primary having different passwords?

haaawk:
Also isn't [this](https://github.com/chiselstrike/turso-cli/pull/391/files#diff-ddf0f1c434b9a822221abc406e7ed9304cb352d372d441d75e25da6dd787302aL65-L70) a breaking change for old databases?

athoscouto:
> Won't it lead to replicas and primary having different passwords?

No, that won't happen. Everything is handled on the server side now.

> Also isn't [this](https://github.com/chiselstrike/turso-cli/pull/391/files#diff-ddf0f1c434b9a822221abc406e7ed9304cb352d372d441d75e25da6dd787302aL65-L70) a breaking change for old databases?

Why would it be a breaking change?

haaawk:
> Why would it be a breaking change?

Ok. If we're handling the password on the server side then it's all good. I wish the PR had mentioned that password handling is moved to server side.

PR #392, opened 2023-05-04
Use new /v1/locations endpoint

athoscouto:
Also refactors how we use `settings`.
Now settings are read only once and only written back after the command is executed, if necessary.


athoscouto:
This does not introduce any functional changes to users, so I'll go ahead and merge it.
If there is anything you want to change, we can address on a following one.

PR #393, opened 2023-05-04
Add cache to isJwtTokenValid and create turso.Tokens client

athoscouto:


internal/turso/turso.go
```diff
@@ -25,6 +25,7 @@ type Client struct {
 	Feedback      *FeedbackClient
 	Organizations *OrganizationsClient
 	Locations     *LocationsClient
+	Tokens        *TokensClient
```
haaawk, line 28: Where is TokensClient defined?

internal/settings/cache.go
```diff
@@ -99,3 +106,19 @@ func (s *Settings) LocationsCache() (map[string]string, string) {
...
+	setCacheWithExp(key, exp, true)
+}
+
+func (s *Settings) TokenValidCache(token string) bool {
+	key := TOKEN_VALID_CACHE_KEY_PREFIX + strings.ReplaceAll(token, ".", "_")
+	ok, err := getCache[bool](key)
+	if err != nil {
+		return false
+	}
+	return ok
```
haaawk, line 123: Ultra nit: This is just `return err != nil && ok`

internal/settings/cache.go
```diff
@@ -23,10 +24,16 @@ func cacheKey(key string) string {
 }
 
 func setCache[T any](key string, ttl int64, value T) error {
-	entry := Entry[T]{Data: value}
+	exp := int64(0)
 	if ttl > 0 {
-		entry.Expiration = time.Now().Unix() + ttl
+		exp = time.Now().Unix() + ttl
 	}
```
haaawk, line 30: Why not just `exp := time.Now().Unix() + ttl`. That would be preserving the previous behaviour. Setting exp to 0 for ttl = 0 changes how this case is handled in getCache.

internal/turso/turso.go
athoscouto: On `internal/turso/token.go`. But it wasn't there. Rebase hell made it end up in the location PR by accident.

internal/settings/cache.go
athoscouto: Will do! But you almost got me there 😬

internal/settings/cache.go
athoscouto: I wanted non-expiring entries as well.  I don't remember why I wanted them.
And we don't have any, so I'll remove this.

haaawk: Currently entry.Expiration == 0 means returning nil in getCache.

haaawk (approved):

PR #394, opened 2023-05-05
Improve DB creation from local sqlite file

athoscouto:
Changes the spinner to inform which task the `turso db create` command currently is doing.
Waits for the database to load the seed, which can take a while as well. 

Also adds a `--wait` or `-w` flag to both `turso db create` and `turso db replicate`, which makes the command wait until the new instance is ready to receive requests.

---

Example of creating a database with a local SQlite file of 43MB:

https://user-images.githubusercontent.com/7853668/236575684-676500cb-ab05-46b2-8cfd-75392c3bebbf.mov

---

How to generate a similar file locally:
```bash
echo "CREATE TABLE blobs (blob BLOB NOT NULL);" > dump.sql
for i in {1..1000}; do echo "INSERT INTO t VALUES (zeroblob(32768));" >> dump.sql; done;
sqlite3 db.sqlite < dump.sql
```
Then just:
```bash
turso db create --from-file db.sqlite
```

### DIFF
--- a/internal/cmd/db_create.go
+++ b/internal/cmd/db_create.go
@@ -17,6 +17,7 @@ func init() {
 	addCanaryFlag(createCmd)
 	addDbFromFileFlag(createCmd)
 	addLocationFlag(createCmd, "Location ID. If no ID is specified, closest location to you is used by default.")
+	addWaitFlag(createCmd, "Wait for the database to be ready to receive requests.")
 }
 
 var createCmd = &cobra.Command{
@@ -104,7 +105,7 @@ var createCmd = &cobra.Command{
 			return err
 		}
 
-		if dbFile != nil {
+		if waitFlag || dbFile != nil {
 			description = fmt.Sprintf("Waiting for database %s to be ready", internal.Emph(name))
 			spinner.Text(description)
 			if err = client.Instances.Wait(name, instance.Name); err != nil {
--- a/internal/cmd/db_create.go
+++ b/internal/cmd/db_create.go
@@ -95,12 +95,21 @@ var createCmd = &cobra.Command{
 				return fmt.Errorf("could not create database %s: %w", name, err)
 			}
 
-			description = fmt.Sprintf("Finishing creating database %s%s in %s ", internal.Emph(name), dbText, internal.Emph(regionText))
+			description = fmt.Sprintf("Finishing to create database %s%s in %s ", internal.Emph(name), dbText, internal.Emph(regionText))
 			spinner.Text(description)
 		}
 
-		if _, err = client.Instances.Create(name, "", region, image); err != nil {
+		instance, err := client.Instances.Create(name, "", region, image)
+		if err != nil {
 			return err
+		}
+
+		if dbFile != nil {
+			description = fmt.Sprintf("Waiting for database %s to be ready", internal.Emph(name))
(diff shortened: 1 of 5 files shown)


penberg:
Looks good to me! @CodingDoug please have a look

CodingDoug:
It looks like the wait flag, as a standalone CLI function, is kind of orthogonal to loading from a file, since specifying a file forces it to wait anyway.  Should we just require the wait flag in all cases where the CLI should wait for readiness, in order to keep its behavior consistent? So the default behavior would be for the CLI to always terminate as quickly as possible, unless it's told to wait.

Also I wonder (and this is very speculative) if waiting should actually be its own operation for the entire database, not just a flag for certain instance operations.  It could work in any circumstance: `turso db wait $DB_NAME`.  That would extend its potential utility to other operations that might cause some downtime (like a database version update), and allows the user to do a followup wait if they didn't think to put the flag in the command in the first place.

athoscouto:
> Should we just require the wait flag in all cases where the CLI should wait for readiness, in order to keep its behavior consistent?

I don't think so, the CLI should feel snappy. The usual DB startup time after creation should be small most of the time (< 5s), so most users won't even query the DB until it is ready.

> So the default behavior would be for the CLI to always terminate as quickly as possible, unless it's told to wait.

Yes, but the problem with creating the DB from a file is that it can take a while. Minutes in some cases. And if we don't wait the user may think their database is broken. 

> Also I wonder (and this is very speculative) if waiting should actually be its own operation for the entire database, not just a flag for certain instance operations.

I don't think so, because it waiting by itself doesn't achieve anything for the user.
I'd rather add the flag to the create and replicate operations than have another command.
For other long operations, like update, it should be the other way around: the default is waiting but you could also use a flag to detach yourself from the execution.

athoscouto:
> I think I was unclear. I'm saying that the CLI should always end as fast as possible unless the wait flag is specified. I'm not saying the CLI should always wait.
 
You were clear, I mixed things up though. 

> That can be addressed with messaging, especially providing a command to run (below) to make sure the database is ready.

Messaging is very good on command descriptions and on docs.
I don't like it on command outputs very much for two reasons:
- The more unstructured text, the harder it is to automate things and reuse commands.
- The CLI feels cluttered, especially if already know your way around it. 

We'll soon have a status endpoint that users can use to check if the DB is ready, up, down, broken, etc.
I'm still not convinced about the wait command though.

> In this case, I'd rather there be consistency in behavior across the board rather than special-casing guesses at what the user probably wants, or how the system will probably behave, or timing on what the user will probably do in response. It's easier to explain the behavior of systems in this way and set expectations accordingly, and accounts for those times when these "probably" statements aren't all satisfied.

I agree with the principle of valuing consistency. We should also be pragmatic about UX.
I don't see the downside in this specific case of making `--from-file` trigger `--wait`. 
The lack of `--wait` though is making people confused.

We should totally revisit it once we have a better story around getting database status/state though.

penberg:
Let's work on the DX in tree.

penberg:
I have no objections to `turso db wait`, but on the other hand I am perfectly happy with `--wait` flag too.

haaawk:
It seems that for creation from a file it would be better to always wait. Just to avoid confusion.

PR #400, opened 2023-05-08
feat: show user plan and limits on account show

### LINKED ISSUE #334: Show account plan and limits in `turso account show`
The CLI is currently hard-coded to the starter plan because historically that was the only plan that existed. Now that we're working towards other plans, let's query the API for account plan and limits.

StefanoSaffran:
This PR aims to update `turso account show` command to use the data from API to show the account plan and limits

closes #334

### DIFF
--- a/internal/cmd/account.go
+++ b/internal/cmd/account.go
@@ -1,6 +1,43 @@
 package cmd
 
-import "github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


haaawk (approved):
Thanks @StefanoSaffran

PR #406, opened 2023-05-09
fix: add validation to avoid api tokens/databases creation with invalid name

### LINKED ISSUE #403: Place limits on valid characters for platform API tokens
I can create a token with nothing but invisible characters:

```
$ go run cmd/turso/main.go auth api-tokens create "   "

$ go run cmd/turso/main.go auth api-tokens list
NAME

```

I now have no easy way of identifying it in the list.

I can also attempt to embed other whitespace characters into the name.  Here are two tabs:

```
$ go run cmd/turso/main.go auth api-tokens create "               "
Error: failed to create token: parse "/v1/auth/api-tokens/\t\t": net/url: invalid control character in URL
exit status 1
```

It would be reasonable to place the same restrictions on token names as the names of databases.

StefanoSaffran:
This PR aims to add a validation to avoid API tokens creation with an invalid name.

closes #403

### DIFF
--- a/internal/cmd/auth_createapitokens.go
+++ b/internal/cmd/auth_createapitokens.go
@@ -1,10 +1,13 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/chiselstrike/iku-turso-cli/internal"
 	"github.com/chiselstrike/iku-turso-cli/internal/prompt"
+	"github.com/chiselstrike/iku-turso-cli/internal/turso"
 	"github.com/spf13/cobra"
 )
 
@@ -27,7 +30,12 @@ var createApiTokensCmd = &cobra.Command{
 			return err
 		}
 
-		tokenName := args[0]
+		tokenName := strings.TrimSpace(args[0])
+
+		if !turso.IsValidName(tokenName) {
+			return errors.New("invalid name: names only support lowercase letters, numbers, and hyphens")
+		}
+
 		description := fmt.Sprintf("Creating api token %s", internal.Emph(tokenName))
 		bar := prompt.Spinner(description)
 		defer bar.Stop()
--- a/internal/cmd/db_create.go
+++ b/internal/cmd/db_create.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"time"
@@ -9,6 +10,7 @@ import (
 	"github.com/chiselstrike/iku-turso-cli/internal"
 	"github.com/chiselstrike/iku-turso-cli/internal/prompt"
 	"github.com/chiselstrike/iku-turso-cli/internal/settings"
+	"github.com/chiselstrike/iku-turso-cli/internal/turso"
 	"github.com/spf13/cobra"
 )
 
@@ -30,6 +32,10 @@ var createCmd = &cobra.Command{
 		name, err := getDatabaseName(args)
 		if err != nil {
 			return err
+		}
+
+		if !turso.IsValidName(name) {
+			return errors.New("invalid name: names only support lowercase letters, numbers, and hyphens")
 		}
 
 		config, err := settings.ReadSettings()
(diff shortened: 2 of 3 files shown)


internal/cmd/auth_createapitokens.go
```diff
@@ -27,7 +29,14 @@ var createApiTokensCmd = &cobra.Command{
 			return err
 		}
 
-		tokenName := args[0]
+		tokenName := strings.TrimSpace(args[0])
+
+		validNameRegex := regexp.MustCompile("^[a-zA-Z0-9_-]+$")
```
athoscouto, line 34: I think we should enforce the same rules for all entities' names.
The ones that we use for databases (on the server) seem good enough:

```go
func IsValidName(name string) bool {
	for _, r := range name {
		if !(unicode.IsDigit(r) || (unicode.IsLetter(r) && unicode.IsLower(r)) || r == '-') {
			return false
		}
	}
	return true
}
```

Let's use the same function here instead of regexes.
And since you're at it, please also validate database names before we send the request to the API.

athoscouto, line 34: It doesn't quite apply here, but in future uses of the `regexp` package please:
- Use `regexp.MustCompile` in a place that is clearly called once at the beginning of the program life (e.g. at the package level) and reused.
- Otherwise use `regexp.Compile`.

internal/cmd/auth_createapitokens.go
```diff
@@ -27,7 +29,14 @@ var createApiTokensCmd = &cobra.Command{
...
+
+		if !validNameRegex.MatchString(tokenName) {
+			fmt.Println("Invalid token name, creation aborted.")
```
athoscouto, line 37: Please return an error to interrupt the program execution here.

haaawk:
This is still a draft @athoscouto ;)

internal/cmd/auth_createapitokens.go
StefanoSaffran: Thanks, I added the function and used it instead of regex.

internal/cmd/auth_createapitokens.go
```diff
@@ -27,7 +29,15 @@ var createApiTokensCmd = &cobra.Command{
 			return err
 		}
 
-		tokenName := args[0]
+		tokenName := strings.TrimSpace(args[0])
+
+		isValid := turso.IsValidName(tokenName)
+
+		if !isValid {
```
athoscouto, line 36: ⛏️
```suggestion
		if !turso.IsValidName(tokenName) {
```

internal/cmd/db_create.go
```diff
@@ -32,6 +33,13 @@ var createCmd = &cobra.Command{
 			return err
 		}
 
+		isValid := turso.IsValidName(name)
+
+		if !isValid {
```
athoscouto, line 38: :pick: 
```suggestion
		if !turso.IsValidName(tokenName) {
```

internal/cmd/db_create.go
```diff
@@ -32,6 +33,13 @@ var createCmd = &cobra.Command{
...
+			fmt.Println("Invalid name format. Please use only lowercase letters, numbers, and hyphens (-).")
+			return nil
```
athoscouto, line 40:
```suggestion
			return errors.New("invalid name: names only support lowercase letters, numbers, and hyphens")
```

internal/cmd/auth_createapitokens.go
```diff
@@ -27,7 +29,15 @@ var createApiTokensCmd = &cobra.Command{
...
+			fmt.Println("Invalid name format. Please use only lowercase letters, numbers, and hyphens (-).")
+			return nil
```
athoscouto, line 38:
```suggestion
			return errors.New("invalid name: names only support lowercase letters, numbers, and hyphens")
```

athoscouto (approved):
LGTM

A couple of nits regarding the `isValid` variable.

Also, we should try to act uniformly during error handling.
In general, we return an `error` instead of printing a message and returning `nil`.
Because [error strings should not be capitalized](https://github.com/golang/go/wiki/CodeReviewComments#error-strings), I've suggested a change in the error message.

haaawk (approved):

PR #407, opened 2023-05-10
fix: abort auth commands when token is set in env

### LINKED ISSUE #270: Handling of auth commands with valid token in TURSO_API_TOKEN env
Assuming the user is already signed in with a token that works:

```
$ turso auth token
[EXISTING TOKEN]

$ export TURSO_API_TOKEN=[EXISTING TOKEN]

$ turso auth logout
Logged out.

$ turso db list
[DATABASE LIST OK]
```

At this point, we have a token in env, nothing in settings.json, and commands requiring auth are working OK.

However, the following sequence of commands don't make sense given this context:

```
$ turso auth logout
No user logged in.
```

Is that response accurate?  I'm implicitly logged in by way of env token.

```
$ turso auth login
https://api.chiseledge.com?port=55346&redirect=true
Waiting for authentication...
✔  Success! Logged in as CodingDoug

$ grep -i token settings.json
  "token": "[NEW TOKEN]",

$ echo $TURSO_API_TOKEN
[EXISTING TOKEN]
```

At this point, there are two tokens in the system, env and settings.json.  Auth login went ahead and did its thing without regard to env.

```
$ turso db list
[DATABASE LIST OK]
```

Which token was responsible for allowing this command to work?  It's unclear.

```
$ turso auth logout
Logged out.

$ turso auth token
Error: no user logged in. Run turso auth login to log in and get a token

$ turso db list
[DATABASE LIST OK]
```

The fact that the command still works after logout could be confusing.  The CLI is now back to using the first token, but we've received no acknowledgement about this.

Proposal:

- If there is a token in env, *all* auth commands should fail saying that auth commands aren't effective when a token is in env.
- If there is a token provided with `-t`, similar to above.

Even better, if the CLI can decode the token, for any command that requires a token:

- If the token can't be parsed at all, error out and say that.
- If the token is expired, say that, and suggest `turso auth login` to get a new one.

StefanoSaffran:
This PR aims to make all `auth` commands fail if the token is set in env.

closes #270

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -26,6 +26,7 @@ var authCmd = &cobra.Command{
 	Use:               "auth",
 	Short:             "Authenticate with Turso",
 	ValidArgsFunction: noSpaceArg,
+	PersistentPreRunE: verifyIfTokenIsSetInEnv,
 }
 
 var signupCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


haaawk (approved):
Thanks @StefanoSaffran

haaawk:
It seems that Github actions have some outage right now. I will try to rerun them tomorrow morning and then merge if they pass.

PR #414, opened 2023-05-16
Improve db list performance

haaawk:


internal/cmd/db_list.go
```diff
@@ -31,29 +35,64 @@ var listCmd = &cobra.Command{
...
+			return err
+		}
+
+		var data [][]string
+		for range databases {
+			data = append(data, <-dbInfo)
 		}
+		sort.Slice(data, func(i, j int) bool {
+			return data[i][0] > data[j][0]
+		})
```
athoscouto, line 95: :pick:: You can use the index of the for loop to populate the data table directly, removing the need for the `dbInfo` channel.

athoscouto (approved):
LGTM

All comments are nits suggesting not using channels, since `errgroup` is doing the synchornization.
Since they don't change functionality, feel free to ignore.

haaawk:
You're obviously correct @athoscouto but I prefer being explicit about the synchronization. Using channels to communicate between goroutines seems a bit more idiomatic than sharing a memory that's implicitly synchronized because of some other primitive doing something else :).

PR #424, opened 2023-05-18
Run actions on self-hosted

haaawk:


haaawk:
Does not seem to work @dyasny

PR #425, opened 2023-05-18
Run Github actions on self-hosted

haaawk:


haaawk:
@dyasny please get the runners unstuck :)

PR #431, opened 2023-05-20
implement hrana v1 for turso dev

glommer:
Also add tests, and fix a few corner cases with the root path

### DIFF
--- a/internal/cmd/dev.go
+++ b/internal/cmd/dev.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"github.com/gin-gonic/gin"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"database/sql"
@@ -20,6 +21,17 @@ func init() {
 	addDevPortFlag(devCmd)
 	addDevFileFlag(devCmd)
 	addVerboseFlag(devCmd)
+}
+
+var statementRegex = regexp.MustCompile(`^\s*(?i)(INSERT|DELETE|UPDATE)\b`)
+
+// need to decide if we'll use Query or Exec
+func shouldExec(statement string, args ...interface{}) bool {
+	matches := statementRegex.FindStringSubmatch(statement)
+	if len(matches) >= 1 {
+		return true
+	}
+	return false
 }
 
 type resultSet struct {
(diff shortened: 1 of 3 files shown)


internal/cmd/dev.go
```diff
@@ -89,16 +85,350 @@ func (s *statement) UnmarshalJSON(data []byte) error {
...
+		for _, val := range values {
+			switch v := val.(type) {
+			case nil:
+				row = append(row, hranaValue{Type: "null"})
+			case int64:
+				row = append(row, hranaValue{Type: "integer", Value: fmt.Sprintf("%d", v)})
+			case float64:
+				row = append(row, hranaValue{Type: "float", Value: v})
+			case []byte:
+				row = append(row, hranaValue{Type: "blob", Value: string(v)})
```
honzasp, line 237: Note that the correct format is `{"type": "blob", "base64": base64-encoded blob}`

honzasp:
Nice! I will have a more detailed look on Monday :) To test this, you can use unit tests of `@libsql/client`, you can run them as `URL='http://localhost:8080' VERSION=1 npm run test` in the `libsql/libsql-client-ts` repo.

internal/cmd/dev.go
```diff
@@ -89,16 +98,390 @@ func (s *statement) UnmarshalJSON(data []byte) error {
...
+
+type hranaBatchSuccessResponse struct {
+	Result hranaBatchResponse `json:"result"`
+}
+
+func convertValue(val hranaValue) (interface{}, error) {
+	if val.Type == "blob" {
+		return []byte(val.Base64), nil
+	} else if val.Type == "integer" {
+		return strconv.Atoi(val.Value.(string))
```
honzasp, line 190: Hrana (and SQLite) integers are 64-bit, but the Go `int` may be 32-bit, so we should use `strconv.ParseInt()` instead of `Atoi()`.

internal/cmd/dev.go
```diff
@@ -89,16 +98,390 @@ func (s *statement) UnmarshalJSON(data []byte) error {
...
+			for _, val := range values {
+				switch v := val.(type) {
+				case nil:
+					row = append(row, hranaValue{Type: "null"})
+				case int64:
+					row = append(row, hranaValue{Type: "integer", Value: fmt.Sprintf("%d", v)})
+				case float64:
+					row = append(row, hranaValue{Type: "float", Value: v})
+				case []byte:
+					row = append(row, hranaValue{Type: "blob", Base64: string(v)})
```
honzasp, line 289: We need to base64-encode the bytes here.

internal/cmd/dev.go
```diff
@@ -89,16 +98,390 @@ func (s *statement) UnmarshalJSON(data []byte) error {
...
+		c.JSON(http.StatusBadRequest, herr)
+		return
+	}
+	result.Result = *res
+	c.JSON(http.StatusOK, result)
+}
+
+// 0 -> skipped
+// 1 -> error
+// 2 -> ok
```
honzasp, line 322: Maybe we can use the Go enum-like syntax with `iota` to define these as named constants?

internal/cmd/dev.go
```diff
@@ -89,16 +98,390 @@ func (s *statement) UnmarshalJSON(data []byte) error {
...
+func evaluateCond(cond hranaCond, stmtRes []int) bool {
+	ty := strings.ToLower(cond.Type)
+	if ty == "ok" {
+		// malformed
+		if cond.Step == nil {
+			return false
+		}
+		step := *cond.Step
+		return stmtRes[step] == 2
```
honzasp, line 331: Passing an out-of-range `step` will cause a panic. Other protocol violations (such as passing invalid `hranaValue`) will also cause a panic, so perhaps we can also panic when `cond.Step == nil`, rather than silently ignoring the error?

internal/cmd/dev.go
```diff
@@ -89,16 +98,390 @@ func (s *statement) UnmarshalJSON(data []byte) error {
...
+	stmtRes := make([]int, len(batch.Steps))
+	var result hranaBatchResponse
+
+	results := make([]*hranaStmtResult, len(batch.Steps))
+	errors := make([]*hranaError, len(batch.Steps))
+
+	for step_idx, step := range batch.Steps {
+		var res *hranaStmtResult
+		var err *hranaError
+		skip := false
```
honzasp, line 386: Nit: the code would be slightly more readable if we used a positive predicate (`enable := true`) instead of a negative (`skip`).

internal/cmd/dev.go
```diff
@@ -89,16 +98,390 @@ func (s *statement) UnmarshalJSON(data []byte) error {
...
+		}
+		return curr
+	}
+
+	return false
+}
+
+func executeHranaBatch(batch hranaBatch, db *sql.DB) hranaBatchResponse {
+	stmtRes := make([]int, len(batch.Steps))
+	var result hranaBatchResponse
```
honzasp, line 378: Nit: Why do we declare this variable now? We can return `hranaBatchResponse{results, errors}` directly.

internal/cmd/dev.go
```diff
@@ -89,16 +98,390 @@ func (s *statement) UnmarshalJSON(data []byte) error {
...
+	}
+
+	if verboseFlag {
+		fmt.Printf("Executing %s, %s\n", internal.Emph(statement), internal.Emph(params))
+	}
+
+	result.Rows = make([][]hranaValue, 0)
+	result.Cols = make([]hranaCol, 0)
+
+	if shouldExec(statement, params...) {
```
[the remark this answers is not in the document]
glommer: yeah, I don't like this, but I didn't find any good parser in go for sqlite. Without a parser, this will never work well. This is starting to change my mind on this approach. Maybe it is indeed better to just spawn sqld.

PR #419, opened 2023-05-18
Fix db hostname not found in shell cmd

Mr-Destructive:
#401 

Have returned an error statement when the function is not able to find the hostname in the list of DBs. 

Let me know if this is a valid fix or if any further changes are required.
Thank you.

PR #435, opened 2023-05-22
call sqld instead

glommer:
This PR calls sqld instead of trying to implement hrana in go.

It puts the burden on the user to install it, but as soon as we start publishing binaries we can download it ourselves.

### DIFF
--- a/internal/cmd/dev.go
+++ b/internal/cmd/dev.go
@@ -1,206 +1,85 @@
 package cmd
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
-	"github.com/gin-gonic/gin"
-	"net/http"
+	"os"
+	"os/exec"
+	"os/signal"
+	"path/filepath"
+	"syscall"
 
-	"database/sql"
 	"github.com/chiselstrike/iku-turso-cli/internal"
 	"github.com/spf13/cobra"
-	"io"
-	_ "modernc.org/sqlite"
 )
 
 func init() {
 	rootCmd.AddCommand(devCmd)
 	addDevPortFlag(devCmd)
 	addDevFileFlag(devCmd)
-	addVerboseFlag(devCmd)
-}
-
-type resultSet struct {
-	Columns []string        `json:"columns"`
-	Rows    [][]interface{} `json:"rows"`
-}
-
-type statementSuccessResponse struct {
-	Results resultSet `json:"results"`
-}
-
-type errorResponse struct {
-	Error struct {
-		Message string `json:"message"`
-	} `json:"error"`
-}
-
-type statement struct {
-	Query  string `json:"q"`
-	Params []interface{}
-}
-
-func (s statement) String() string {
-	var buf bytes.Buffer
-	buf.WriteString(s.Query)
-	buf.WriteString("; ")
-	jsonData, err := json.Marshal(s.Params)
-	if err != nil {
-		panic("can't convert json data from params")
-	}
-	buf.WriteString(string(jsonData))
-	return buf.String()
-}
-
-type statementList struct {
-	Statements []statement `json:"statements"`
-}
-
-func (s *statement) UnmarshalJSON(data []byte) error {
-	var query interface{}
-	err := json.Unmarshal(data, &query)
-	if err != nil {
-		return err
-	}
-
-	switch q := query.(type) {
-	case string:
-		s.Query = q
-	case map[string]interface{}:
-		qValue, ok := q["q"].(string)
-		if !ok {
-			return fmt.Errorf("invalid JSON: 'q' field is missing or not a string")
-		}
-		s.Query = qValue
-
-		switch p := q["params"].(type) {
-		case []interface{}:
-			s.Params = p
-		case map[string]interface{}:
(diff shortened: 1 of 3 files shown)


glommer:
Will close #431 if you all think this is a better approach (@CodingDoug @honzasp )

CodingDoug:
I like pushing as much responsibility down to the actual software rather than a simulation when possible, without adding much dev overhead.

I'm wondering though, at this point, what value is the CLI adding over just running sqld yourself? I'm sure there is room to improve going forward, but I want to be clear what exactly we intend to offer in the near term.

glommer:
At this point the main thing is that we make sure the addresses are set up correctly, and provide better visibility into the sqlite file. The user can use a visible file in the local directory of their choosing, that I then link into the data/ directory of sqld, that is temporary.

A good next step is to download sqld once we have builds, and follow ups, as you suggested, are auto-downloading the schema and matching versions of databases, installing extensions, etc.

honzasp:
> btw, one thing we can do very soon - if someone helps me with that, is add certificates so we can keep using libsql:// URLs.

I'm not sure how we could do that, we have no control over the list of trusted CAs in all clients, and no CA will give us a certificate for `localhost:8080`. But I'm planning to add support for `?tls=0` query string to libsql client (after the new HTTP API is implemented), then we could use `libsql://localhost:8080?tls=0`.

glommer:
@CodingDoug I'd advocate for merging this, since it already provides one QoL feature, which is the ability of passing a local sqlite file (and then we'll handle links internally), and we can soon install sqld automatically. But would love to hear from you

CodingDoug:
@glommer Sure, let's do whatever we can do make things easier in the near term.

glommer:
do you want to test drive first?

@haaawk this should also fix the build on windows, as it removes moderc.org/sqlite

CodingDoug:
> do you want to test drive first?

Yes, I'll do this shortly.

CodingDoug:
# Command output

I'd suggest changing this output:

```
For environments where a filesystem is available:
→   use file:somefile.db as a URL. This server is not needed. (Ctrl-C)
For all others environments:
→   connect to ws://127.0.0.1:8080.
```

Without context, it's confusing how I should act on this messaging. If my intent in running `turso dev` was to start a local server, I'm not clear why it would be trying to talk me out of that.

What I suggest is changing the help text for this command to the following:

> $ turso dev --help
> Starts a sqld emulator for local development with Turso.
>
> If you're using a libSQL client SDK that supports SQLite database files on the local filesystem, then you might not need this server at all.  Instead, you can use a `file:` URL with the path to the file you want the SDK to read and write.

The change the normal output:

> $ turso dev
> sqld emulator listening on port 8080. Use the this URL to configure your libSQL client SDK for local development:
>
>     ws://127.0.0.1:8080
> 
> [persistence-message]

[persistence-message] can be either:

> This server is using an in-memory database. Changes will be lost when this server stops. If you want to persist changes, use `--db-file` to specify a SQLite database file instead.

or:

> Using database file `[db-file]`

# Command line args

Suggest using `--db-file` with short flag `-f` instead of `--local-file`.  Files are implicitly local for the developer. 

Suggest adding `-p` as a short flag for `--port`.

`--verbose` doesn't appear to do anything.  Minimally I'd like to see a record of each incoming request so I know it's working, maybe even each JSON payload.

# turso db shell

The shell does not work with 127.0.0.1 or localhost URLs.  This is unexpected. It errors with:

> Error: hostname 'localhost' not found

I thought this used to work for running against other local sqld. Maybe something bad happened in #419?

glommer:
I'll also remove verbose for now. We won't have the necessary level of control over the sqld verbosity level without patching it, so that's for later, best case.  All we can do at the moment is increase the rust log level from error to info, debug or trace, but that will print. *a lot* of unrelated stuff.

glommer:
I don't think the issues with localhost are really solvable from our side.
It seems the client is always trying the ipv6 address on the mac, even though the hosts file lists both.
the shell issue was fixed.

glommer:
Ok, to support both localhost and 127.0.0.1 on mac, we'd have to patch sqld to accept more than one listen-addr for the server, and explicitly listen on both 127.0.0.1:8080 and `[::1]:8080`.

If I listen on the latter, then localhost works (but 127.0.0.1 doesn't)

CodingDoug:
127.0.0.1 works fine for me. We can assume the user will copy the provided URL, and the server of course do whatever it's coded to do (listen on 127.0.0.1).

internal/cmd/dev.go
```diff
@@ -1,206 +1,85 @@
...
+	"os/exec"
+	"os/signal"
+	"path/filepath"
+	"syscall"
 
-	"database/sql"
 	"github.com/chiselstrike/iku-turso-cli/internal"
 	"github.com/spf13/cobra"
-	"io"
-	_ "modernc.org/sqlite"
```
haaawk, line 14: We probably need `go mod tidy` to remove unused dependencies.

glommer: should we add to CI a test that requires go mod tidy to come empty?

PR #438, opened 2023-05-23
Add --from-file flag to turso db shell

athoscouto:
This will enable users to load dumps into Turso.
Since `turso db create --from-file` will fail for big enough files.

I created https://github.com/libsql/libsql-shell-go/issues/90 and https://github.com/libsql/libsql-shell-go/issues/91 to address the shortcomings found while implementing this PR.

### DIFF
--- a/internal/cmd/auth_createapitokens.go
+++ b/internal/cmd/auth_createapitokens.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 
 	"github.com/chiselstrike/iku-turso-cli/internal"
-	"github.com/chiselstrike/iku-turso-cli/internal/prompt"
+	"github.com/chiselstrike/iku-turso-cli/internal/prompt/spinner"
 	"github.com/chiselstrike/iku-turso-cli/internal/turso"
 	"github.com/spf13/cobra"
 )
@@ -37,7 +37,7 @@ var createApiTokensCmd = &cobra.Command{
 		}
 
 		description := fmt.Sprintf("Creating api token %s", internal.Emph(tokenName))
-		bar := prompt.Spinner(description)
+		bar := spinner.Start(description)
 		defer bar.Stop()
 
 		data, err := client.ApiTokens.Create(tokenName)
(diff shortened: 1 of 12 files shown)


glommer:
@athoscouto most database shells have an `-f` switch, btw, to execute statements from a file. This is generally useful beyond the initial scope because not always you want to create a database from a file. Could that work here as well ?

haaawk (approved):

CodingDoug:
> @athoscouto most database shells have an `-f` switch, btw, to execute statements from a file. This is generally useful beyond the initial scope because not always you want to create a database from a file. Could that work here as well ?

@glommer I think stdin should work as well so people can pipe without having to drop tmp files.  Technically, we do support stdin, but it's slow because the shell is reading it through a "cooked" tty instead of accepting raw data.  The shell could instead check to see if the input is coming from a tty and handle it differently.

Also see: #365

glommer:
stdin is obviously fine, and should work better once we improve some stuff. 

But see for instance:
```
> psql --help | grep '\-f'
  -f, --file=FILENAME      execute commands from file, then exit
```

Although checking now, mysql doesn't have it, nor does sqlite3. So maybe it's more of a postgres thing, than a "common way"

CodingDoug:
@glommer  When you say "most shells", you are maybe thinking of `awk` and `sed` and the like. :)

CodingDoug:
> The problem with stdin is that, by default, it executes one command at a time.

Certainly stdin can be treated as a normal stream just like any file-based stream if it's detected to be a tty?  There should be nothing requiring it to be "cooked" as a line-based input mechanism.  This related issue discusses that: https://github.com/chiselstrike/turso-cli/issues/365#issuecomment-1534705891

luisfvieirasilva (approved):

glommer:
I think we should be very careful with things that are user-visible. If the existence of the flag is supposed to be temporary in nature, I'd avoid merging.

That said, postgres has this and I always found `-f` useful, so... The batch behavior we can fix later as long as it is not exposed.

athoscouto:
I'll close this one because https://github.com/libsql/sqld/pull/423 will enable us to do it without batches.
I agree with @CodingDoug points around pipes.
I think the `-f` flag is still valuable though.
Files are not just streams, we can open them, get stats, and even print % of progress.

PR #442, opened 2023-05-23
Accepts shell connections from unkown hostnames

### LINKED ISSUE #440: Turso shell no longer recognizes localhost or 127.0.0.1 URLs
I'm pretty certain that the shell used to work with sqld running locally, but it no longer does. This might have broken with #419. This was a great feature to have, and will be important also to work with `turso dev` in the future (see #435).

luisfvieirasilva:
This is useful for sqld running locally.

ps: A revert of merge commit a9bc7c60b72b44c8a104a4898e40c3a0521da653

Closes #440

### DIFF
--- a/internal/cmd/db_shell.go
+++ b/internal/cmd/db_shell.go
@@ -150,7 +150,7 @@ func databaseFromURL(dbURL string, client *turso.Client) (*turso.Database, error
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


glommer:
ack

glommer (approved):

glommer:
@haaawk why is CI not running here?

PR #444, opened 2023-05-24
make sure the tree is always tidy

glommer:
includes a run of go mod tidy to clean existing issues


haaawk (approved):

glommer:
@haaawk this seems to be failing for unrelated reasons

PR #446, opened 2023-05-25
Add --enable-extensions to turso db create

athoscouto:
It allows users to enable extensions.
Usage:
```
➜  iku-turso-cli git:(athos/experimental) ✗ turso db create --help
Create a database.

Usage:
  turso db create [flags] [database_name]

Flags:
      --canary              Use database canary build.
      --enable-extensions   Enables experimental support for SQLite extensions.
                            The following extensions and extension sets are currently included:
                                - sqlean
                                - sqlite-vss
                            If you would like to see some extension included, run turso account feedback.
                            Warning: extensions support is experimental and subject to change
      --from-file string    create the database from a local SQLite3-compatible file
  -h, --help                help for create
      --location string     Location ID. If no ID is specified, closest location to you is used by default.
  -w, --wait                Wait for the database to be ready to receive requests.

Global Flags:
  -c, --config-path string                  Path to the directory with config file
      --no-multiple-token-sources-warning   Don't warn about multiple access token sources
```
      
```
turso db create --enable-extensions
```

### DIFF
--- a/internal/cmd/canary_flag.go
+++ b/internal/cmd/canary_flag.go
@@ -2,8 +2,8 @@ package cmd
 
 import "github.com/spf13/cobra"
 
-var canary bool
+var canaryFlag bool
 
 func addCanaryFlag(cmd *cobra.Command) {
-	cmd.Flags().BoolVar(&canary, "canary", false, "Use database canary build.")
+	cmd.Flags().BoolVar(&canaryFlag, "canary", false, "Use database canary build.")
 }
--- a/internal/cmd/db_create.go
+++ b/internal/cmd/db_create.go
@@ -17,6 +17,7 @@ import (
 func init() {
 	dbCmd.AddCommand(createCmd)
 	addCanaryFlag(createCmd)
+	addEnableExtensionsFlag(createCmd)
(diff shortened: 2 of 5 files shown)


glommer:
I would call it "--experimental-features", otherwise users can have the impression that basic functionality may not be available unless they pass that

athoscouto:
I'd rather keep the flag name simple and use the flag usage (already done) and docs to convey that info.
Mainly for two reasons:

- Keeping names simple improves overall CLI UX.
- I don't think this flag will be only used for experimental stuff. We may decide to have opt-in features (e.g. geospatial). 

WDYT?

glommer:
tagging @CodingDoug as a tie breaker but I don't like the name "features", as I said. Could be --extra-features, or anything else.

athoscouto:
What do you all think about `--enable extensions`?

CodingDoug:
Then what happens if there are multiple different features to enable or disable? If you give each feature its own flag, it doesn't require much extra cognitive load to understand how they work.

athoscouto:
Then you just do `--enable feature1 --enable feature2` or `enable=feature1,feature2`.

PR #458, opened 2023-06-01
show latency from your location

glommer:
Probe latencies and show them at db locations

### DIFF
--- a/internal/cmd/db.go
+++ b/internal/cmd/db.go
@@ -2,7 +2,9 @@ package cmd
 
 import (
 	"fmt"
+	"math"
 	"os"
+	"sync"
 
 	"github.com/chiselstrike/iku-turso-cli/internal"
 	"github.com/chiselstrike/iku-turso-cli/internal/settings"
@@ -127,8 +129,56 @@ func getHost() string {
 	return host
 }
 
+type latMap struct {
+	id  string
+	lat int
+}
+
 func locations(client *turso.Client) (map[string]string, error) {
 	settings, _ := settings.ReadSettings()
+	return readLocations(settings, client)
+}
+
+func latencies(client *turso.Client) (map[string]int, error) {
+	settings, _ := settings.ReadSettings()
+	locations, err := readLocations(settings, client)
+	if err != nil {
+		return nil, err
+	}
+
+	var wg sync.WaitGroup
+	latencies := make(map[string]int)
+	c := make(chan latMap, len(locations))
+	for id := range locations {
+		wg.Add(1)
+		go func(id string) {
+			defer wg.Done()
+			measure := math.MaxInt
+			// XXX: Running this in different goroutines makes all latencies dogslow.
+			// Not sure if this is contention at the client or API level
+			for i := 0; i < 3; i++ {
+				d := turso.ProbeLocation(id)
+				if d != nil {
+					measure = int(math.Min(float64(d.Milliseconds()), float64(measure)))
+				}
+			}
+			if measure != math.MaxInt {
+				c <- latMap{id: id, lat: measure}
+			}
+
+		}(id)
+	}
+
+	wg.Wait()
+	close(c)
+
+	for kvp := range c {
+		latencies[kvp.id] = kvp.lat
+	}
+	return latencies, nil
+}
+
+func readLocations(settings *settings.Settings, client *turso.Client) (map[string]string, error) {
 	if locations := settings.LocationsCache(); locations != nil {
 		return locations, nil
 	}
--- a/internal/cmd/db_locations.go
+++ b/internal/cmd/db_locations.go
@@ -2,15 +2,19 @@ package cmd
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 
 	"github.com/chiselstrike/iku-turso-cli/internal"
+	"github.com/rodaine/table"
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/maps"
+	"unicode/utf8"
 )
 
 func init() {
 	dbCmd.AddCommand(regionsCmd)
+	addLatencyFlag(regionsCmd)
 }
 
 var regionsCmd = &cobra.Command{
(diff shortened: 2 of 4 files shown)


glommer:
@CodingDoug this PR - if accepted - will show an estimated location from you to each of the locations we support.

I don't know if we want an option to disable it or not, seems to me it should be always on

internal/cmd/db.go
```diff
@@ -127,8 +130,64 @@ func getHost() string {
...
+				measure := int(kvp.lat.Milliseconds())
+				lat, ok := latencies[kvp.id]
+				if ok {
+					latencies[kvp.id] = int(math.Min(float64(lat), float64(measure)))
+				} else {
+					latencies[kvp.id] = measure
+				}
+			}
+		}
+	}()
```
athoscouto, line 181: Please do this synchronously or synchronize it through the wait group.

internal/cmd/utils.go
```diff
@@ -59,6 +59,23 @@ func filterInstancesByRegion(instances []turso.Instance, region string) []turso.
...
+
+	start := time.Now()
+	_, err = client.Do(req)
+	if err != nil {
+		return nil
+	}
+	dur := time.Since(start)
+	return &dur
+}
+
```
athoscouto, line 78: ⛏️: Please move this into `internal/turso/locations.go` client.

internal/cmd/utils.go
```diff
@@ -59,6 +59,23 @@ func filterInstancesByRegion(instances []turso.Instance, region string) []turso.
 	return result
 }
 
+func probeLocation(location string) *time.Duration {
+	client := &http.Client{}
+	req, err := http.NewRequest("GET", "http://turso-bouncer.fly.dev:8080/", nil)
```
athoscouto, line 64: The "official" URL should be `http://region.turso.io:8080`. My bad for giving you the internal one.

athoscouto (approved):
I'm only worried about the synchronization comment. The rest we can clean up later.

internal/cmd/db.go
glommer: why are you worried? It's a single writer.

But will change.

athoscouto: Because there is a race condition between you writing the results to the map and writing the map to the cache.
I may have misunderstood the code though.

glommer:
@athoscouto let's give @CodingDoug a chance to comment before merge.

penberg:
@glommer The `turso db locations` command seems to now just hang on my machine with this pull request.

glommer:
Best explanation I have is some requests hang for you for some reason. I can add a timeout -- which is a good idea anyway - and the failed ones should show a ??? for you, and at least help us debug it.

glommer:
@penberg try now ?

glommer:
nevermind, it hangs for me too. It has to do with the change I made late to move the map construction out of a goroutine as athos requested. Investigating

glommer:
@athoscouto with the loop to retrieve the values after the Wait, the channel fills up =( We can add a channel with higher capacity, but my understanding is that this is an anti-pattern, because as we add more locations, and/or more retries, the channel has to be updated accordingly.

There is no race condition in retrieving the values in a go routine, as far as I can see. There is a single retriever. `close()` happens after `Wait`, making sure that the channel is only closed after all latency measurements were taken.

internal/cmd/db.go
```diff
@@ -127,8 +130,64 @@ func getHost() string {
...
+		return nil, err
+	}
+
+	if latencies := settings.LatencyCache(); latencies != nil {
+		return latencies, nil
+	}
+
+	var wg sync.WaitGroup
+	latencies := make(map[string]int)
+	c := make(chan latMap)
```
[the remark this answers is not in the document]
haaawk: ok you actually don't have to but it would allow goroutines from lines 160-165 to finish independently and release resources sooner.

glommer:
That's doable, but what's wrong with a goroutine that consumes as results are produced? Athos claims it is racy, I don't see a race here.

(btw, tests seem to be failing for some unrelated reason)

internal/cmd/db.go
```diff
@@ -127,8 +130,64 @@ func getHost() string {
...
+				defer wg.Done()
+				dur := turso.ProbeLocation(id)
+				c <- latMap{id: id, lat: dur}
+
+			}(id)
+		}
+	}
+
+	go func() {
+		for kvp := range c {
```
haaawk, line 170: With buffered channel in line 156, this code could be in the main function after wg.Wait in line 183

glommer: but what would be the main advantage?

internal/cmd/db.go
```diff
@@ -127,8 +130,64 @@ func getHost() string {
...
+				} else {
+					latencies[kvp.id] = measure
+				}
+			}
+		}
+	}()
+
+	wg.Wait()
+	close(c)
+	settings.SetLatencyCache(latencies)
```
haaawk, line 185: I agree with @athoscouto that this code can execute in parallel to the goroutine from lines 169-181

CodingDoug:
First observations:

The latency column is initially a bit overwhelming to take in. I think that's because the numbers are all over the place, and they are not sorted. This column would make a whole lot more sense if sorted.  The prior alphabetical sorting *almost* seems useless now because it's not helping me find what appears to be the most important items in the list.  In fact, I now realize the old sorting wasn't doing me any real good at all since it was sorting on the code and not a human-relatable name.

I have a concern also that there could be a discrepancy between the recommended default location and the best observed latency. If the latency is measured by the CLI but the recommended location is measured by Fly, that could create some conflict that has no obvious resolution.  Should the user pick the default or the lowest latency?  Or are the latency measurements fundamentally transient and not a long-term indicator of best performance for me?

Also a bit of a concern: we aren't explaining how the latency is calculated, so it could be misinterpreted as some indicator of overall general load at each location, which could make further locations appear unappealing as potential replica locations. The user has to already be thinking about network topology specifically from the perspective of the machine where the command is currently being run in order to make sense of this output.

CodingDoug:
I'm thinking the latency column would be easier to read if the values were right-aligned, so that the numbers are easier to compare with each other.

Also I'm wondering if maybe the entire column should be optional behind a CLI flag. You can feel the weight of this addition when running old vs. new. And I'm concerned that flakey networks might make this command feel especially sluggish if we have to wait for all these values to come in.

glommer:
@CodingDoug do you feel like it adds value, behind a cli flag, or just doesn't add value at all ?

CodingDoug:
> @CodingDoug do you feel like it adds value, behind a cli flag, or just doesn't add value at all ?

It adds value, but I think context is necessary to make it easier to understand what's being shown and understand how to make decisions based on it.

I will say that if it doesn't ultimately help someone make a decision, then it's both overhead and noise by default.

glommer:
Then maybe `db locations` is not the place to show it?

For the record, I have coded this as a way to show this information somewhere else, in an upcoming PR. We just figured the latencies would be nice to show by default on locations, but if you disagree we don't have to.

(about sluggishness, the latencies are cached, so if you run again it should be fast)

CodingDoug:
> Then maybe `db locations` is not the place to show it?

This command makes sense. I just question it as default output, which is why I wanted to consider a CLI flag.

> For the record, I have coded this as a way to show this information somewhere else, in an upcoming PR. We just figured the latencies would be nice to show by default on locations, but if you disagree we don't have to.

I draw a fairly thick line between "nice to have" and "helps me make decisions" :)

It all boils down to fulfilling expected use cases. If the data doesn't address a common use case, then it might be considered noise. Hiding it behind a flag make this all good, and also gives an opportunity to:

- Explain in the flag help text what this does
- Sort the output by latency

> (about sluggishness, the latencies are cached, so if you run again it should be fast)

I saw the cache, but that only really works for the cache duration, which is not that long (15 minutes?). I don't see someone really needing to run this command repeatedly within that time. I consider the main use cases of `db locations` to be:

- Help me understand the locations I have available (and the best location choice for me) when I first get started.
- Help me understand the locations I have available for adding a replica

These cases only require a single invocation, and (I imagine) are not highly likely to occur in the same 15 minute window.

CodingDoug:
Love it. If you can right-align the latency column data, even better.

glommer:
Unfortunately I didn't find any way to right align it in the table
implementation I am using. That bugs me too ;(

Of course we can do a pass on the strings and pad with spaces ourselves...
or change table implementation.

I will take a quick look

glommer:
It will be a bit challenging and error prone to right align this, especially without a very good understanding how the table package works. (They do have an open bug asking for a way to right align)

I'm merging as is and we can revisit later

PR #464, opened 2023-06-05
show a table if the user does not pick a location when replicating

glommer:
Andreas (chronark) points out that DX could be better if the user is presented with the list of locations right away. We will also take the opportunity to teach them that it is possible to pass a --location next time.

### DIFF
--- a/internal/cmd/db_replicate.go
+++ b/internal/cmd/db_replicate.go
@@ -2,8 +2,10 @@ package cmd
 
 import (
 	"fmt"
+	"sort"
 	"time"
 
+	tbl "github.com/charmbracelet/bubbles/table"
 	"github.com/chiselstrike/iku-turso-cli/internal"
 	"github.com/chiselstrike/iku-turso-cli/internal/prompt"
 	"github.com/chiselstrike/iku-turso-cli/internal/settings"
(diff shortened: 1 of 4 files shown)


glommer:
@CodingDoug for your review.

internal/cmd/db_replicate.go
```diff
@@ -29,29 +31,77 @@ func replicateArgs(cmd *cobra.Command, args []string, toComplete string) ([]stri
...
+}
+
 var replicateCmd = &cobra.Command{
 	Use:               "replicate database_name location_id",
 	Short:             "Replicate a database.",
-	Args:              cobra.RangeArgs(2, 3),
+	Args:              cobra.RangeArgs(1, 3),
 	ValidArgsFunction: replicateArgs,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		client, err := createTursoClientFromAccessToken(true)
```
haaawk, line 72: This should be after cmd.SilenceUsage = true

internal/cmd/db_replicate.go
```diff
@@ -29,29 +31,77 @@ func replicateArgs(cmd *cobra.Command, args []string, toComplete string) ([]stri
 	return dbNameArg(cmd, args, toComplete)
 }
 
+func pickLocation(locations map[string]string, exclude []string) string {
+	fmt.Printf("Pro-tip! Next time, you can add the %s parameter.\n", internal.Emph("--location"))
```
haaawk, line 35: location in `db replicate` is not a flag but a positional argument.

internal/cmd/db_replicate.go
glommer: The problem is that we want the check for args being valid to be before that, no ?

That creates a circularity, since we need the client to show the table.

CodingDoug:
This table (and likely the other location table introduced recently) need to be able to scroll its contents in order to accommodate shorter terminal windows.  If the window is shorter than the list (by some margin), the user can't see the topmost options and the selection cursor. This is obviously problematic.

The intro text is also part of this issue. Consider if there is intro text, and it's not allowed to stay in view along with the table, it's effectively invisible for shorter windows.

glommer:
have to, no. But it's certainly nice, no?

I think in practice this will work for almost any terminal size, plus this is contained to helper paths.

glommer:
( I just force-pushed a version that actually fixes the table size )

CodingDoug:
> have to, no. But it's certainly nice, no?

Fancy, yes.  Bu tnice can mean different things to different people.  For a CLI, my objective judgement of "nice" is that it gets me to accomplish a task as fast as possible, which is why I mention number of keystrokes as a metric to optimize. Tapping the down arrow 20+ times to get to yyz is not "nice".

glommer:
@CodingDoug I added code to the main table implementation (so should cover all interactive tables) that lets you search the id list by typing. So if you press "y" you should go to Montreal (yul).

How about that?

glommer:
(it's a bit buggy and will need another pass, but just as a PoC)

PR #497, opened 2023-06-22
Feat/db restore

luisfvieirasilva:
Add new `turso db restore {db_name} -o {restore_path}` command

### DIFF
--- a/internal/cmd/db_restore.go
+++ b/internal/cmd/db_restore.go
@@ -1,0 +1,114 @@
+//go:build preview
+// +build preview
+
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"time"
+
+	"github.com/chiselstrike/iku-turso-cli/internal"
+	"github.com/chiselstrike/iku-turso-cli/internal/prompt"
+	"github.com/chiselstrike/iku-turso-cli/internal/turso"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	dbCmd.AddCommand(dbRestoreCmd)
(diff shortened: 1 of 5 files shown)


internal/cmd/db_restore.go
```diff
@@ -0,0 +1,114 @@
...
+			return fmt.Errorf("please specify a database name")
+		}
+
+		timestamp, err := getTimestamp(cmd)
+		if err != nil {
+			return err
+		}
+		if outputFlag == "" {
+			return fmt.Errorf("please specify an output file")
+		}
```
athoscouto, line 43: Let's add a sensible default, let's say `restore.db`?

internal/cmd/timestamp_flag.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+var timestampS string
+var fromNow time.Duration
+
+func addTimestampFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&timestampS, "timestamp", "", "Set timestamp target (RFC3339 format)")
+	cmd.Flags().DurationVar(&fromNow, "from-now", time.Duration(0), "Set timestamp target relative from now")
```
athoscouto, line 15:
```suggestion
	cmd.Flags().StringVar(&timestampS, "timestamp", "", "Set timestamp for point in time restore (RFC3339 format)")
	cmd.Flags().DurationVar(&fromNow, "from-now", time.Duration(0), "Set timestamp for point in time restore, relative to the current moment")
```

athoscouto (approved):

PR #498, opened 2023-06-23
switch to organization as we create it

glommer:
Once you create an organization, the assumption is that you want to switch to it.

### DIFF
--- a/internal/cmd/org.go
+++ b/internal/cmd/org.go
@@ -21,6 +21,51 @@ func init() {
 	membersCmd.AddCommand(membersRemoveCmd)
 }
 
+func switchToOrg(client *turso.Client, slug string) error {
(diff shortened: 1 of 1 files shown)


internal/cmd/org.go
```diff
@@ -21,6 +21,51 @@ func init() {
...
+		return err
+	}
+
+	if org.Type == "personal" {
+		slug = ""
+	}
+
+	settings.SetOrganization(slug)
+
+	fmt.Printf("Default organization set to %s.\n", internal.Emph(org.Slug))
```
CodingDoug, line 62: This should be "current" rather than "default".  The CLI doesn't really have a sense of default because you can't override it in any way.  It must be set as the current choice.

PR #499, opened 2023-06-23
better status code if organization already exists

glommer:
409 is not that obvious

### DIFF
--- a/internal/turso/organizations.go
+++ b/internal/turso/organizations.go
@@ -52,6 +52,10 @@ func (c *OrganizationsClient) Create(name string) (Organization, error) {
 	}
 	defer r.Body.Close()
 
+	if r.StatusCode == http.StatusConflict {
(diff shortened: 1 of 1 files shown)


internal/turso/organizations.go
```diff
@@ -52,6 +52,10 @@ func (c *OrganizationsClient) Create(name string) (Organization, error) {
 	}
 	defer r.Body.Close()
 
+	if r.StatusCode == http.StatusConflict {
+		return Organization{}, fmt.Errorf("failed to create organization %s: name already exists", internal.Emph(name))
```
athoscouto, line 56: tip: `parseResponseError` can be useful for extracting error messages from HTTP responses.
Even though we can craft better messages when we know what error happened, we forget to handle some individual cases, so using `parseResponseError` catch-all is a plus.

athoscouto (approved):

PR #500, opened 2023-06-27
Fix `account show` for scaler plan

penberg:
We do need to move these to the API, but let's at least show the right values when on the scaler plan.

### DIFF
--- a/internal/cmd/account_show.go
+++ b/internal/cmd/account_show.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"strconv"
 
 	"github.com/chiselstrike/iku-turso-cli/internal"
 	"github.com/dustin/go-humanize"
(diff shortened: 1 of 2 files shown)


internal/cmd/account_show.go
```diff
@@ -49,34 +48,30 @@ var accountShowCmd = &cobra.Command{
...
+		addResourceRowCount(tbl, "locations", usage.Total.Locations, planInfo.maxLocations)
 		tbl.Print()
 
 		return nil
 	},
 }
 
 func addResourceRowBytes(tbl table.Table, resource string, used, limit uint64) {
+	if limit == 0 {
+		tbl.AddRow(resource, "Unlimited", "Unlimited", "")
```
haaawk, line 64: Wouldn't it be good to showed used anyway?

internal/cmd/account_show.go
```diff
@@ -49,34 +48,30 @@ var accountShowCmd = &cobra.Command{
...
+		return
+	}
 	tbl.AddRow(resource, humanize.IBytes(used), humanize.IBytes(limit), percentage(float64(used), float64(limit)))
 }
 
 func addResourceRowCount(tbl table.Table, resource string, used, limit uint64) {
+	if limit == 0 {
+		tbl.AddRow(resource, "Unlimited", "Unlimited", "")
```
haaawk, line 72: Wouldn't it be good to showed used anyway?

haaawk (approved):

PR #512, opened 2023-06-30
fix shell issues when connecting to specific URLs

glommer:
There are two issues when connecting to specific URLs that are fixed by this PR:

1. When connecting to non-turso URLs (like localhost running sqld), we will try to get turso tokens first, which will obviously fail. The code depends on db objects very early, so the best way is to test if an URL is turso.io (or not an URL), and only try to get to tokens in that case.

2. When connecting to specific DB instances, the current code is mistakenly overriding the dbURL with the URL we get from the DB name. That's just plain wrong, and is fixed here.

### DIFF
--- a/internal/cmd/db.go
+++ b/internal/cmd/db.go
@@ -17,7 +17,6 @@ var showHttpUrlFlag bool
 var showInstanceUrlsFlag bool
 var showInstanceUrlFlag string
 var yesFlag bool
-var instanceFlag string
 
 func getInstanceNames(client *turso.Client, dbName string) []string {
 	instances, err := client.Instances.List(dbName)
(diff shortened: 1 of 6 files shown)


CodingDoug:
It might be out of the scope of this PR, but I'll say that what I'd really like to see is the following options for shell:

- `turso db shell DBNAME`
- `turso db shell DBNAME --instance INSTANCENAME`
- `turso db shell URL`

The first two can assume that they are Turso databases that require an auth token.  The second can assume that it is a non-Turso DB without auth that can only be referenced by URL. This might simplify things greatly, and make it easier for people to connect Turso instances without having to get a URL for it (which they would normally never need to do anyway).

glommer:
I find it a bit unnatural to specify an instance name instead of location here. But in general it is a good suggestion

glommer:
I have noticed that other commands like `db destroy` will take both location and instance, so I added this here too

PR #515, opened 2023-07-03
Cache `db list` response (fixes #502)

### LINKED ISSUE #502: turso db shell needs to list databases
The latency to talk to Turso API can be an order of magnitude slower than connecting to a Turso DB.

We should cache the result of turso db list in a similar way that we cache names.
The logic may need to be a little bit more intricate, so we don't risk consistency.

avinassh:
The idea is to save the `client.Databases.List()` response locally. At first, I thought I could just save this struct, however this introduces a cyclic dependency between `settings` and `turso` module. So I created a new struct which resembles it and added helper methods to convert between them.

I also renamed caching methods like `GetDbNamesCache` to `GetDatabasesCache` since we aren't storing the names anymore, but full metadata. 

I did keep the same key name in the config `database_names` otherwise that key would remain unused in the file. (side note: I think similar thing happened to `cached_db_names` as I can't find any references to it in the codebase)

--- 

Another alternative would be moving the `turso.Database` to a common place, but that would require more changes and I also felt that is unnecessary.

### DIFF
--- a/internal/cmd/db.go
+++ b/internal/cmd/db.go
@@ -66,14 +66,6 @@ func convertInternalDbToCachedDb(databases []turso.Database) []settings.Database
 	return dbs
 }
 
-func fetchDatabaseNames(client *turso.Client) []string {
-	databases, err := client.Databases.List()
-	if err != nil {
-		return []string{}
-	}
-	return extractDatabaseNames(databases)
-}
-
 func getDatabase(client *turso.Client, name string) (turso.Database, error) {
 	databases, err := client.Databases.List()
 	if err != nil {
--- a/internal/cmd/db.go
+++ b/internal/cmd/db.go
@@ -38,6 +38,34 @@ func extractDatabaseNames(databases []turso.Database) []string {
 	return names
 }
 
+func convertCachedDbToInternalDb(databases []settings.Database) []turso.Database {
+	dbs := make([]turso.Database, 0)
+	for _, database := range databases {
+		dbs = append(dbs, turso.Database{
+			ID:            database.ID,
+			Name:          database.Name,
+			Regions:       database.Regions,
+			PrimaryRegion: database.PrimaryRegion,
+			Hostname:      database.Hostname,
+		})
+	}
+	return dbs
+}
+
+func convertInternalDbToCachedDb(databases []turso.Database) []settings.Database {
+	dbs := make([]settings.Database, 0)
+	for _, database := range databases {
+		dbs = append(dbs, settings.Database{
+			ID:            database.ID,
+			Name:          database.Name,
+			Regions:       database.Regions,
+			PrimaryRegion: database.PrimaryRegion,
+			Hostname:      database.Hostname,
+		})
+	}
+	return dbs
+}
+
 func fetchDatabaseNames(client *turso.Client) []string {
 	databases, err := client.Databases.List()
 	if err != nil {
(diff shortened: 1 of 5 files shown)


internal/settings/cache.go
```diff
@@ -52,23 +52,33 @@ func invalidateCache[T any](key string) error {
...
 }
 
-func (s *Settings) GetDbNamesCache() []string {
-	data, err := getCache[[]string](DB_NAMES_CACHE_KEY)
+func (s *Settings) SetDatabasesCache(dbNames []Database) {
+	setCache(DB_CACHE_KEY, DB_CACHE_TTL_SECONDS, dbNames)
+}
+
+func (s *Settings) GetDatabasesCache() []Database {
+	data, err := getCache[[]Database](DB_CACHE_KEY)
```
haaawk, line 73: Will this work with existing cache that contains string values not Database values?

internal/cmd/db.go
```diff
@@ -38,6 +38,34 @@ func extractDatabaseNames(databases []turso.Database) []string {
 	return names
 }
 
+func convertCachedDbToInternalDb(databases []settings.Database) []turso.Database {
+	dbs := make([]turso.Database, 0)
```
haaawk, line 42: why not `make([]turso.Database, len(databases))`?

internal/cmd/db.go
```diff
@@ -38,6 +38,34 @@ func extractDatabaseNames(databases []turso.Database) []string {
...
+		dbs = append(dbs, settings.Database{
+			ID:            database.ID,
+			Name:          database.Name,
+			Regions:       database.Regions,
+			PrimaryRegion: database.PrimaryRegion,
+			Hostname:      database.Hostname,
+		})
+	}
+	return dbs
+}
```
haaawk: I'm wondering whether we couldn't have a single generic function instead of two copies of the same code?

internal/settings/cache.go
avinassh: Yup, it works.

athoscouto (commented):
This type of struct replication problem is one we'll face every time we want to cache a struct.
It seems a good candidate for a general solution.

Have you considered exposing the raw cache functions`getCache`, `setCache`, and `invalidateCache`?
This way the database module could pick the cache key and type, avoiding the cyclic dependency.
This design is a bit cleaner because the caching layer becomes agnostic and doesn't have to know about the application.

We would probably want a higher-level version of `GetCache` that already handles errors and returns zero value in case there is nothing in the cache.

internal/cmd/db.go
```diff
@@ -61,18 +81,28 @@ func getDatabase(client *turso.Client, name string) (turso.Database, error) {
 	return turso.Database{}, fmt.Errorf("database %s not found. List known databases using %s", internal.Emph(name), internal.Emph("turso db list"))
 }
 
-func getDatabaseNames(client *turso.Client) []string {
+func getDatabasesCacheOrAPI(client *turso.Client) ([]turso.Database, error) {
```
athoscouto, line 84: ⛏️: nit about naming. `getDatabases` is just fine. Callers shouldn't be worried about where the data is coming from, they should just have an way to get the value.

penberg:
@avinassh Nice! There were some nits/suggestions from @athoscouto and @haaawk. If they make sense, please just send a new pull request to address them incrementally.

PR #519, opened 2023-07-04
feat: add debug flag to show outgoing http requests dumps

gris:
Added a `--debug` flag to all commands in order to show dumps of outgoing http requests.

Example:

Screen Shot 2023-07-04 at 10 59 51 AM

### DIFF
--- a/internal/turso/turso.go
+++ b/internal/turso/turso.go
@@ -6,8 +6,11 @@ import (
 	"io"
 	"mime/multipart"
 	"net/http"
+	"net/http/httputil"
 	"net/url"
 	"os"
+
+	"github.com/chiselstrike/iku-turso-cli/internal/flags"
 )
 
 // Collection of all turso clients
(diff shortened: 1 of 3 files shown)


internal/turso/turso.go
```diff
@@ -79,6 +82,13 @@ func (t *Client) newRequest(method, urlPath string, body io.Reader) (*http.Reque
 
 func (t *Client) do(method, path string, body io.Reader) (*http.Response, error) {
 	req, err := t.newRequest(method, path, body)
+	if flags.Debug() {
+		dump, err := httputil.DumpRequestOut(req, true)
+		fmt.Printf("%s", string(dump))
```
athoscouto, line 87:
```suggestion
		fmt.Println(string(dump))
```

Or just `fmt.Print` if you don't want the extra line break.

internal/turso/turso.go
```diff
@@ -79,6 +82,13 @@ func (t *Client) newRequest(method, urlPath string, body io.Reader) (*http.Reque
 
 func (t *Client) do(method, path string, body io.Reader) (*http.Response, error) {
 	req, err := t.newRequest(method, path, body)
+	if flags.Debug() {
```
athoscouto, line 85: Please make sure to check the error before you do anything with the return value.

internal/turso/turso.go
athoscouto, line 87: Please make sure to check the error before you do anything with the return value.

internal/turso/turso.go
```diff
@@ -79,6 +82,13 @@ func (t *Client) newRequest(method, urlPath string, body io.Reader) (*http.Reque
...
+		dump, err := httputil.DumpRequestOut(req, true)
+		fmt.Printf("%s", string(dump))
+		if err != nil {
+			return nil, err
```
athoscouto, line 89: Let's not break the CLI because the dump functionality is not working properly! 😬
Just print the error and move on.

athoscouto (approved):

PR #524, opened 2023-07-04
fix: add user agent header

gris:
Should we delete the header `TursoCliVersion`?

closes https://github.com/chiselstrike/turso-cli/issues/517

### DIFF
--- a/internal/turso/turso.go
+++ b/internal/turso/turso.go
@@ -77,7 +77,7 @@ func (t *Client) newRequest(method, urlPath string, body io.Reader) (*http.Reque
 		req.Header.Add("Authorization", fmt.Sprint("Bearer ", t.token))
 	}
 	req.Header.Add("TursoCliVersion", t.cliVersion)
-	req.Header.Add("User-Agent", fmt.Sprintf("Turso CLI/%s (%s/%s)", t.cliVersion[1:], runtime.GOOS, runtime.GOARCH))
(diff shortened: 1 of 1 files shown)


CodingDoug:
If we do, I think it should conform to the typical user agent format.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

> User-Agent: <product> / <product-version> <comment>

gris:
Thanks for bringing that to my awareness. Changed it, would love to hear comments about the new format :)

internal/turso/turso.go
```diff
@@ -76,6 +77,7 @@ func (t *Client) newRequest(method, urlPath string, body io.Reader) (*http.Reque
 		req.Header.Add("Authorization", fmt.Sprint("Bearer ", t.token))
 	}
 	req.Header.Add("TursoCliVersion", t.cliVersion)
+	req.Header.Add("User-Agent", fmt.Sprintf("Turso CLI/%s (%s/%s)", t.cliVersion[1:], runtime.GOOS, runtime.GOARCH))
```
athoscouto, line 80: ⛏️: WDYT about using the repo name?
```suggestion
	req.Header.Add("User-Agent", fmt.Sprintf("turso-cli/%s (%s/%s)", t.cliVersion[1:], runtime.GOOS, runtime.GOARCH))
```

athoscouto (approved):

internal/turso/turso.go
gris: lgtm!

Issue #479, opened 2023-06-20
Transferring databases to an organization


penberg:
One use case is that people create a database in their personal workspace. They then create an organization and want to transfer the database there. Let's add a `turso org transfer` or `turso db transfer` command to support this.

PR #529, opened 2023-07-05
feat: add db transfer command

gris:
closes https://github.com/chiselstrike/turso-cli/issues/479

### DIFF
--- a/internal/cmd/db_transfer.go
+++ b/internal/cmd/db_transfer.go
@@ -11,7 +11,7 @@ import (
 )
 
 func init() {
-	dbCmd.AddCommand(dbTransferCmd)
+	orgCmd.AddCommand(dbTransferCmd)
 }
 
 var dbTransferCmd = &cobra.Command{
@@ -32,7 +32,7 @@ var dbTransferCmd = &cobra.Command{
 			return err
 		}
 
-		ok, err := promptConfirmation(fmt.Sprintf("Are you sure you want to transfer database %s to %s?", dbName, orgName))
+		ok, err := promptConfirmation(fmt.Sprintf("Are you sure you want to transfer database %s to organization %s?", internal.Emph(dbName), internal.Emph(orgName)))
 		if err != nil {
 			return fmt.Errorf("could not get prompt confirmed by user: %w", err)
 		}
@@ -48,23 +48,22 @@ var dbTransferCmd = &cobra.Command{
 
 func transfer(client *turso.Client, dbName, orgName string) error {
 	config, err := settings.ReadSettings()
+	config.InvalidateDatabasesCache()
 
 	if err != nil {
 		return err
 	}
 
-	msg := fmt.Sprintf("Transfering database %s to org %s", internal.Emph(dbName), internal.Emph(orgName))
+	msg := fmt.Sprintf("Transferring database %s to organization %s", internal.Emph(dbName), internal.Emph(orgName))
 	s := prompt.Spinner(msg)
 	defer s.Stop()
 
 	if err := client.Databases.Transfer(dbName, orgName); err != nil {
-		return fmt.Errorf("error transfering database")
+		return fmt.Errorf("error transferring database")
 	}
 
 	s.Stop()
-	fmt.Printf("✔  Success! Database %s transferred successfully to org %s\n", internal.Emph(dbName), internal.Emph(orgName))
-
-	config.InvalidateDatabasesCache()
+	fmt.Printf("✔  Success! Database %s transferred successfully to organization %s\n", internal.Emph(dbName), internal.Emph(orgName))
 
 	return nil
 }
--- a/internal/cmd/db_transfer.go
+++ b/internal/cmd/db_transfer.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/chiselstrike/iku-turso-cli/internal"
 	"github.com/chiselstrike/iku-turso-cli/internal/prompt"
-	"github.com/chiselstrike/iku-turso-cli/internal/settings"
 	"github.com/chiselstrike/iku-turso-cli/internal/turso"
 	"github.com/spf13/cobra"
 )
@@ -15,7 +14,7 @@ func init() {
 }
 
 var dbTransferCmd = &cobra.Command{
-	Use:               "transfer database_name org_name",
+	Use:               "db-transfer database_name org_name",
 	Short:             "Transfers database to another organization",
 	Args:              cobra.ExactArgs(2),
 	ValidArgsFunction: dbNameAndOrgArgs,
@@ -31,8 +30,11 @@ var dbTransferCmd = &cobra.Command{
 		if _, err := getDatabase(client, dbName); err != nil {
 			return err
 		}
+		fmt.Printf("To transfer %s database to another organization, all its replicas must be updated.\n", internal.Emph(dbName))
+		fmt.Printf("All your active connections to the DB will be dropped and there will be a short downtime.\n\n")
 
 		ok, err := promptConfirmation(fmt.Sprintf("Are you sure you want to transfer database %s to organization %s?", internal.Emph(dbName), internal.Emph(orgName)))
+
 		if err != nil {
 			return fmt.Errorf("could not get prompt confirmed by user: %w", err)
 		}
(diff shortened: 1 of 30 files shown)


internal/cmd/db_transfer.go
```diff
@@ -0,0 +1,70 @@
...
+
+func init() {
+	dbCmd.AddCommand(dbTransferCmd)
+}
+
+var dbTransferCmd = &cobra.Command{
+	Use:               "transfer database_name org_name",
+	Short:             "Transfers database to another organization",
+	Args:              cobra.ExactArgs(2),
+	ValidArgsFunction: dbNameArg,
```
athoscouto, line 21: This will make autocomplete suggest DB name for both arguments.
If you're going to add a `ValidArgsFunction` you'll probably need to call `dbNameArg` for the first argument and something similar for orgs on the second argument.

internal/cmd/db_transfer.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/chiselstrike/iku-turso-cli/internal"
+	"github.com/chiselstrike/iku-turso-cli/internal/prompt"
+	"github.com/chiselstrike/iku-turso-cli/internal/settings"
+	"github.com/chiselstrike/iku-turso-cli/internal/turso"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	dbCmd.AddCommand(dbTransferCmd)
```
athoscouto, line 14: This will be rarely used. I'd rather not give it the same relevant as `db create`, `db replicate` and so on.

Have you considered adding it under `turso org`?
Something like `turso org db-transfer`.

@penberg thoughts?

internal/cmd/db_transfer.go
```diff
@@ -0,0 +1,70 @@
...
+			return err
+		}
+		dbName := args[0]
+		orgName := args[1]
+
+		if _, err := getDatabase(client, dbName); err != nil {
+			return err
+		}
+
+		ok, err := promptConfirmation(fmt.Sprintf("Are you sure you want to transfer database %s to %s?", dbName, orgName))
```
athoscouto, line 35: ⛏️: WDYT?
```suggestion
		ok, err := promptConfirmation(fmt.Sprintf("Are you sure you want to transfer database %s to organization %s?", internal.Emph(dbName), internal.Emph(orgName)))
```

internal/cmd/db_transfer.go
```diff
@@ -0,0 +1,70 @@
...
+		return transfer(client, dbName, orgName)
+	},
+}
+
+func transfer(client *turso.Client, dbName, orgName string) error {
+	config, err := settings.ReadSettings()
+
+	if err != nil {
+		return err
+	}
```
athoscouto, line 54: Let's invalidate the cache before, just in case.

internal/cmd/db_transfer.go
```diff
@@ -0,0 +1,70 @@
...
+
+	msg := fmt.Sprintf("Transfering database %s to org %s", internal.Emph(dbName), internal.Emph(orgName))
```
athoscouto, line 56:
```suggestion
	msg := fmt.Sprintf("Transfering database %s to organization %s", internal.Emph(dbName), internal.Emph(orgName))
```

internal/cmd/db_transfer.go
```diff
@@ -0,0 +1,70 @@
...
+	s := prompt.Spinner(msg)
+	defer s.Stop()
+
+	if err := client.Databases.Transfer(dbName, orgName); err != nil {
+		return fmt.Errorf("error transfering database")
+	}
+
+	s.Stop()
+	fmt.Printf("✔  Success! Database %s transferred successfully to org %s\n", internal.Emph(dbName), internal.Emph(orgName))
```
athoscouto, line 65:
```suggestion
	fmt.Printf("✔  Success! Database %s transferred successfully to organization %s\n", internal.Emph(dbName), internal.Emph(orgName))
```

athoscouto (commented):
Have some comments, will let @haaawk and @penberg take a look at it tomorrow as well.

internal/cmd/db_transfer.go
gris: Looks good to me! Semantically I thought it would be better at first glance to be a subcmd of db, but the usability perspective definitely has more weight, I think.

internal/cmd/db_transfer.go
gris: looks cool

internal/cmd/db_transfer.go
gris: fixed

internal/cmd/db_transfer.go
```diff
@@ -0,0 +1,69 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/chiselstrike/iku-turso-cli/internal"
+	"github.com/chiselstrike/iku-turso-cli/internal/prompt"
+	"github.com/chiselstrike/iku-turso-cli/internal/settings"
+	"github.com/chiselstrike/iku-turso-cli/internal/turso"
```
CodingDoug, line 9: Should these be https://github.com/chiselstrike/turso-cli ? iku-turso-cli is old and was renamed to turso-cli.

internal/turso/databases.go
```diff
@@ -199,6 +201,34 @@ func (d *DatabasesClient) Update(database string) error {
...
+	}
+	r, err := d.client.Post(url, bodyReader)
+	if err != nil {
+		return fmt.Errorf("failed to transfer database")
+	}
+	defer r.Body.Close()
+
+	if r.StatusCode == http.StatusForbidden {
+		return fmt.Errorf("user is not an owner of the organizations involved in the transfer")
+	}
```
CodingDoug, line 223: This error message could be more useful if it said which organization the user was not an owner of (assuming that the requirement is that the user is the owner of *both* of the orgs).

internal/cmd/db_transfer.go
gris: The code itself is working fine as is, the old name is there bc the mod wasn't renamed. Might be a good opportunity to do it now. I'll try to do it, and see if everything works fine

internal/turso/databases.go
gris: Yes! Now it is better
Screen Shot 2023-07-06 at 4 04 46 PM

gris: It is using the response of the API to provide this error msg instead of building it in the command line application

CodingDoug:
If run this with a junk org name, it proceeds to ask me to do something that could never work:

```
$ go run cmd/turso/main.go org transfer my-db adf
Are you sure you want to transfer database my-db to organization adf? [y/n]:
```

Suggest checking the org for both presence and ownership before prompting.

internal/cmd/db_transfer.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	orgCmd.AddCommand(dbTransferCmd)
+}
+
+var dbTransferCmd = &cobra.Command{
+	Use:               "db-transfer database_name org_name",
+	Short:             "Transfers database to another organization",
```
athoscouto, line 18:
```suggestion
	Short:             "Transfers a database to another organization",
```

internal/cmd/utils.go
```diff
@@ -257,6 +257,18 @@ func dbNameArg(cmd *cobra.Command, args []string, toComplete string) ([]string,
 	return []string{}, cobra.ShellCompDirectiveNoFileComp
 }
 
+func dbNameAndOrgArgs(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	client, err := createTursoClientFromAccessToken(false)
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveNoSpace
+	}
+	if len(args) == 1 {
+		orgs, _ := client.Organizations.List()
```
athoscouto, line 266: This has the ability to spam our API hard.
Making an HTTP request to offer a completion can also make the CLI feel unresponsive.

Please open an issue to cache this request, similarly to what we do with the databases list.

internal/turso/databases.go
```diff
@@ -207,6 +209,34 @@ func (d *DatabasesClient) Update(database string) error {
...
+		return fmt.Errorf("failed to transfer database")
+	}
+	defer r.Body.Close()
+	if r.StatusCode == http.StatusForbidden {
+		err = parseResponseError(r)
+		return fmt.Errorf("%v", err)
+	}
+
+	if r.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed to transfer database to org %s", org)
```
athoscouto, line 234: What happens when the transfer succeeds but the update fails?

athoscouto (approved):
I'll leave approved but the last comment may be something we need to address, depending on the answer.

internal/turso/databases.go
gris: The inspect and plan data will be dummy. I will add a comment asking for updating it

athoscouto: Let's pretend it is atomic on the CLI.
Let's open an API issue to make it atomic.

PR #535, opened 2023-07-07
Bugfix: Refresh cache on DB list API call

### LINKED ISSUE #531: CLI data cache gets out of date with changes from other sources, need mechanism to force a cache reset
The CLI caches various bits of data from the underlying platform API. When there are changes from other sources, the cache doesn't reset until expiration. This can be confusing for users of the CLI who aren't aware that it does caching at all. Recovering from this before expiration requires manually manipulating settings.json, which is not great. We could provide a way for people to either bypass the cache on a per-command basis, or reset it entirely with a command.

For example: `turso db list --force`
Or: `turso cache reset`

avinassh:
closes #531 

I introduced a bug in #515 where I was caching the DB list API response. This PR fixes the issue and adds some more changes: 

- `db list` call always hits the API and refreshes the cache
- other commands which use db details, hit the cache first 
- made the caching layer generic and moved the helper methods to `cmd` module

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -102,8 +102,7 @@ func isJwtTokenValid(token string) bool {
 	if len(token) == 0 {
 		return false
 	}
-	settings, _ := settings.ReadSettings()
-	if settings.TokenValidCache(token) {
+	if tokenValidCache(token) {
 		return true
 	}
 	client, err := tursoClient(token)
@@ -114,7 +113,7 @@ func isJwtTokenValid(token string) bool {
 	if err != nil {
 		return false
 	}
-	settings.SetTokenValidCache(token, exp)
+	setTokenValidCache(token, exp)
 	return true
 }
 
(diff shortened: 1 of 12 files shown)


avinassh:
@athoscouto not sure why the tests are failing. They are failing locally too, when I ran it on `main`

athoscouto:
It works just fine in main.
From main:
```
> turso db shell dear-bug 
Connected to dear-bug at https://dear-bug-athoscouto.turso.io

Welcome to Turso SQL shell!

Type ".quit" to exit the shell and ".help" to list all available commands.

→  .quit
```

From your branch:
```
> turso db shell dear-bug 
Error: failed to connect to database. err: failed to execute SQL: SELECT 1;
```

avinassh:
I figured out the issue, it was happening because the mapstructure was not able to marshal the data back to struct properly (fixed in https://github.com/chiselstrike/turso-cli/pull/535/commits/2d3763d471009a8ba395864d72a8eba9acec0c99). I added a few tests for caching layer now.

So I went into digging how it worked on my main branch locally and I learned that integration tests require that binary already exists and I was not compiling the binary before switching branches.

athoscouto (approved):
LGTM

PR #542, opened 2023-07-07
Handle parsing errors in settings file

### LINKED ISSUE #533: Empty (or invalid) settings.json yields unhelpful error message
Something wiped out my settings.json and left an empty file there (no idea what happened), but now every time I run the CLI, for simple commands, I get the message:

```
$ turso db list
Error: could not read local settings
```

This is very problematic because it doesn't say where that file lives (could vary by platform), what the problem is with it, or how to recover.  Ideally the output should name the file, the error, and make a recommendation to delete the file if it got corrupted.

athoscouto:
Fixes #533.

Example of output when there is a syntax error:
```
> turso auth login
Warning: could not parse JSON config from file /Users/athoscouto/git/turso-cli/settings.json
Fix the syntax errors on the file, or use the --override-config flag to replace it with a fresh one.
Error: could not retrieve local config: While parsing config: invalid character '\n' in string literal
```

How to fix:
```
> turso auth login --override-config
http://127.0.0.1:8080?port=63258&redirect=true
Waiting for authentication...
✔  Success! Logged in as athoscouto
✏️  We are so happy you are here! Now that you are authenticated, it is time to create a database:
        turso db create
```


CodingDoug:
How about `--reset-config`? Override sounds like what `-c` does today.

athoscouto:
Thanks Doug! "Override" wasn't a good name indeed.
I went with recreate though, let me know if you think reset sounds clearer.

avinassh (approved):

avinassh:
Both `reset-config` or `recreate-config` are fine, but I would prefer former as it is faster to type

PR #547, opened 2023-07-11
fix: delete db when user cancel prompt

gris:
We do not want dangling resources


internal/cmd/db_create.go
```diff
@@ -307,6 +308,7 @@ func handleInstanceCreationError(client *turso.Client, name, locationId string,
 
 	_, locationId, err := promptSelect.Run()
 	if err != nil {
+		client.Databases.Delete(name)
```
athoscouto, line 311: When you return the error right after this call, it will trigger a second delete - or won't it?

gris: whoops that is true, removed the one inside the function

athoscouto (approved):

Issue #522, opened 2023-07-04
Improve debug flag


gris:
- [ ] Dump response as well
- [ ] redact auth token
- [ ] Add token to an env var for debugging purposes

avinassh:
What do you think of adding an env variable like `DEBUG=true` as one more option? This way, you don't have to pass `--debug` every time and you can use commands from history too, quickly.

CodingDoug:
This was one of my recommendations from last week as an option *instead of* using the CLI flag, but I don't see a reason not to offer it *in addition to* the CLI flag.

I wouldn't use `DEBUG` though, as that might conflict with other things in the host shell. It makes sense for all the CLI env vars to have some common prefix, such as `TURSO_`.

PR #550, opened 2023-07-13
Dump http response if debug is enabled

avinassh:
Fixes one of the #522

### DIFF
--- a/internal/turso/turso.go
+++ b/internal/turso/turso.go
@@ -113,7 +113,7 @@ func dumpRequest(req *http.Request) {
 func dumpResponse(req *http.Response) {
 	dump, err := httputil.DumpResponse(req, true)
 	if err != nil {
-		fmt.Printf("Failed to dump the HTTP request, you can either remove the debug flag or ignore this error: %s", err.Error())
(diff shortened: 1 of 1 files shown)


internal/turso/turso.go
```diff
@@ -107,6 +110,15 @@ func dumpRequest(req *http.Request) {
 	fmt.Println(string(dump))
 }
 
+func dumpResponse(req *http.Response) {
+	dump, err := httputil.DumpResponse(req, true)
+	if err != nil {
+		fmt.Printf("Failed to dump the HTTP request, you can either remove the debug flag or ignore this error: %s", err.Error())
```
gris, line 116: ⛏️ : ```suggestion
		fmt.Printf("Failed to dump the HTTP response, you can either remove the debug flag or ignore this error: %s", err.Error())
```

gris (approved):

athoscouto:
Please dump request and response at the same time.
Like this you would dump multiple requests before printing their response, which can get confusing. 

Also add an extra new line after responses, so people can see better where one requests ends and another one begins.

Plan show cmd is a good one to test this.

avinassh:
oof beaten by concurrency bug. Fixed now.

athoscouto (approved):

PR #572, opened 2023-07-19
feat: smooth create org flow asking for payment method and change plan to scaler

gris:
Screen Shot 2023-07-26 at 3 18 33 PM

### DIFF
--- a/internal/cmd/db_replicate.go
+++ b/internal/cmd/db_replicate.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"sort"
 	"time"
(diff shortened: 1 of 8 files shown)


athoscouto:
Let's try to reuse as much as we can strings from the upgrade flow:
Screenshot 2023-07-28 at 10 53 27 AM

My suggestion of messages:
```
➜  api git:(main) turso org create foo
Organizations are only supported in paid plans.

You need to add a payment method before you can create organization foo on the scaler plan.
For information about Turso plans pricing and features, access: https://turso.tech/pricing

Want to add a payment method right now? [y/n]: y

Opening your browser at:
https://billing.turso.io/p/session/test_YWNjdF8xS1U2cEFMWnl2c0E1RUlqLF9PTFlDWXUyY2ZTTHJ6VUhQMDJ0RWFNSmhISmtBRm5q0100aztia3N5

Payment method added successfully.
You can manage your payment methods with turso org billing.

You're creating organization foo on the scaler plan.
Do you want to continue? [y/n]: y

Created organization foo on the scaler plan.
Current organization set to foo.
All your turso commands will be executed in that organization context.
To switch back to your previous organization:

        turso org switch athoscouto

➜  turso-cli git:(main) ✗ 
```

Please add an extra line in the end, the layout looks weird when we end with a line beginning with a tab.

Let's make sure that we are consistent in regard to capitalization and punctuation. 
This means that lines should start with upper case letters. 
And that they should end with punctuation unless it ends with a URL or command.

gris:
Did the changes, except reusing as much as I could from the upgrade flow. It would involve a lot of refactor that I didn't see much value in doing now. But created an issue for it, that would be happy to tackle in the near future. https://github.com/tursodatabase/turso-cli/issues/587

athoscouto (approved):

Issue #578, opened 2023-07-26
Implement some indicator of the version of sqld used by `turso dev`


CodingDoug:
Today it's not easy to know what version of sqld is being used with `turso dev`.  It would be good to have a `--version` flag for that, and also something in the message on stdout when it's launched.

avinassh:
fixed in #580

PR #580, opened 2023-07-27
add sqld version on dev cmd

Mr-Destructive:
For #578 
Have added a flag for getting the `sqld` version from the `dev` command. 
It can be used as `turso dev --version`, also the `dev` command will print the sqld `version` in the prompt amongst the other details.

```bash
$ ./turso dev --version
sqld sqld 0.17.1 (ac41bce6 2023-07-27)

$ ./turso dev 
Using sqld version:sqld sqld 0.17.1 (ac41bce6 2023-07-27)

sqld listening on port 8080.
Use the following URL to configure your libSQL client SDK for local development:

    http://127.0.0.1:8080

No auth token is required when sqld is running locally.

This server is using an ephemeral database. Changes will be lost when this server stops.
If you want to persist changes, use --db-file to specify a SQLite database file instead.
```

Is there any specific way this is to be implemented? If am I wrong in this approach, please let me know, I am happy to work on it.

Some obvious things I would like to know more about:
- Parsing only the version number as `0.17.1`?
- Formatting of the output in a particular way?

Thank you

### DIFF
--- a/internal/cmd/dev.go
+++ b/internal/cmd/dev.go
@@ -16,6 +16,7 @@ func init() {
 	rootCmd.AddCommand(devCmd)
 	addDevPortFlag(devCmd)
 	addDevFileFlag(devCmd)
+	getSqldVersion(devCmd)
 }
 
 var devCmd = &cobra.Command{
@@ -26,6 +27,12 @@ var devCmd = &cobra.Command{
 	ValidArgsFunction: noFilesArg,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		cmd.SilenceUsage = true
+
+		versionFlag, _ := cmd.Flags().GetBool("version")
+		if versionFlag {
+			sqldVersion()
+			return nil
+		}
 
 		tempDir, err := os.MkdirTemp("", "*tursodev")
 		if err != nil {
@@ -44,6 +51,9 @@ var devCmd = &cobra.Command{
 				return fmt.Errorf("Error creating link to file: %w", err)
 			}
 		}
+
+		fmt.Printf("Using sqld version: ")
+		sqldVersion()
 
 		addr := fmt.Sprintf("0.0.0.0:%d", devPort)
 		conn := fmt.Sprintf("http://127.0.0.1:%d", devPort)
--- a/internal/cmd/dev.go
+++ b/internal/cmd/dev.go
@@ -16,7 +16,7 @@ func init() {
 	rootCmd.AddCommand(devCmd)
 	addDevPortFlag(devCmd)
 	addDevFileFlag(devCmd)
-	getSqldVersion(devCmd)
+	addDevSqldVersionFlag(devCmd)
 }
 
 var devCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


CodingDoug:
I'm fine with this as-is.

internal/cmd/dev_sqld_version.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/spf13/cobra"
+)
+
+func getSqldVersion(cmd *cobra.Command) {
```
haaawk, line 11: Please rename to `addDevSqldVersionFlag`

internal/cmd/dev_sqld_version.go
```diff
@@ -0,0 +1,24 @@
...
+	cmd.Flags().BoolP("version", "v", false, "sqld version")
+}
+
+func sqldVersion() {
```
haaawk, line 15: Please make this function return version and error instead of printing them.

internal/cmd/dev_sqld_version.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/spf13/cobra"
+)
+
+func getSqldVersion(cmd *cobra.Command) {
+	cmd.Flags().BoolP("version", "v", false, "sqld version")
```
haaawk, line 12: Please bind the flag to a variable the same way it's done in https://github.com/tursodatabase/turso-cli/blob/main/internal/cmd/dev_port_flag.go and use that variable instead of calling `cmd.Flags().GetBool("version")`

internal/cmd/dev.go
```diff
@@ -45,6 +52,9 @@ var devCmd = &cobra.Command{
 			}
 		}
 
+		fmt.Printf("Using sqld version: ")
+		sqldVersion()
```
haaawk, line 56: Why is the version printed twice? Once only when the flag is set and once no matter what the flag value is.

haaawk (changes requested):
LGTM in general but I left some comments.

internal/cmd/dev.go
Mr-Destructive: Before logging the `sqld` is running at the port, I wanted to show the version used. It's OK not to include there.

Mr-Destructive:
Have made changes as reviewed, let me know if I misinterpreted something.
Thank you.

internal/cmd/dev_sqld_version.go
```diff
@@ -8,17 +8,19 @@ import (
...
+	cmd.Flags().BoolVarP(&sqldVersion, "version", "v", false, "sqld version")
 }
 
-func sqldVersion() {
+func getSqldVersion() (string, error) {
 	sqld := exec.Command("sqld", "--version")
 	sqld.Env = append(os.Environ(), "RUST_LOG=error")
 	version, err := sqld.Output()
 	if err != nil {
 		fmt.Println("Error running sqld --version:", err)
```
haaawk, line 22: I think we can drop this line.

haaawk (approved):
LGTM. Thanks @Mr-Destructive. One last comment and we can merge the PR.

haaawk (approved):

PR #597, opened 2023-08-09
Remove install dependencies step

athoscouto:


.github/workflows/go.yml
```diff
@@ -4,34 +4,25 @@
...
+    branches:
+      - '*'
+  
 
 env:
   TURSO_API_TOKEN: ${{ secrets.TURSOCI_TOKEN }}
 
 jobs:
-
-  authorize:
```
haaawk, line 15: @athoscouto we need this to be able to run CI for external contributions. Or have you found another way of doing that safely?

.github/workflows/go.yml
```diff
@@ -4,34 +4,25 @@
...
-    needs: authorize
     runs-on: self-hosted
     steps:
     - uses: actions/checkout@v3
       with:
         ref: ${{ github.event.pull_request.head.sha || github.ref }}
         #ref: "refs/pull/${{ github.event.number }}/merge" # THIS IS INSECURE
 
-    - name: Setup Golang with cache
-      uses: magnetikonline/action-golang-cache@v4
```
haaawk, line 34: Why are you removing the cache? Isn't that going to make the workflow slower?

.github/workflows/go.yml
```diff
@@ -40,9 +31,6 @@ jobs:
     - name: Format
       run: if [ "$(gofmt -s -l . | wc -l)" -gt 0 ]; then exit 1; fi
 
-    - name: Install dependencies
```
haaawk, line 43: I don't remember exactly but it might have been the case that this step allowed the cache to cache the dependencies better

haaawk (changes requested):
What are you trying to do @athoscouto ? What's the problem with authorize, install dependencies and cache?

athoscouto:
As we discussed offline, this was just a test to fix the failing badge.
I haven't gone anywhere with it, so I'll just close it.
Sorry for the trouble!

PR #603, opened 2023-08-24
Use TURSO_CONFIG_FOLDER env var to change the configuration folder

athoscouto:
This allows us to do something like:
```
export TURSO_CONFIG_FOLDER=~/.turso/prod
```
To switch between local and prod configs and avoid the constant `turso auth logout && turso auth login` while developing/operating the platform.


CodingDoug:
How does this work with the existing -c flag?

athoscouto:
The `-c` flag takes precedence

PR #609, opened 2023-09-04
add message to alert upgrades on personal orgs

gris:
This is an attempt to alert users that they are upgrading personal orgs and try to nudge them into creating team orgs, lots of users are updating their personal accounts by mistake, and this is generating manual work to fix it.

### DIFF
--- a/internal/cmd/plan.go
+++ b/internal/cmd/plan.go
@@ -204,6 +204,18 @@ func ChangePlan(client *turso.Client, plans []turso.Plan, current string, hasPay
 		printPricingInfoDisclaimer()
 	}
 
+	settings, err := settings.ReadSettings()
+	if err != nil {
+		return err
+	}
+
+	org := settings.Organization()
+	if change == "upgrading" && org == "" {
+		fmt.Println("You're currently using your personal organization.")
+		fmt.Println("In personal organizations, you cannot invite members to collaborate.")
+		fmt.Printf("If you want instead to upgrade to scaler plan, you can create an organization with %s.\n", internal.Emph("turso org create"))
+	}
+
 	if ok, _ := promptConfirmation("Do you want to continue?"); !ok {
 		fmt.Printf("Plan change cancelled. You're still on %s.\n", internal.Emph(current))
 		return nil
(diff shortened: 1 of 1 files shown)


athoscouto (commented):
How it is today:

```
You need to add a payment method before you can upgrade to the scaler plan.
For information about Turso plans pricing and features, access: https://turso.tech/pricing

Want to add a payment method right now? [y/n]: y

Opening your browser at:
https://billing.turso.io/p/session/test_YWNjdF8xS1U2cEFMWnl2c0E1RUlqLF9PYVVvdndTZEpxY1lOMHdPS2xoamRZbEloZ21KSWRO0100UpTZYMz9

Payment method added successfully.
You can manage your payment methods with turso org billing.

You're upgrading to the scaler plan.
You're currently using your personal organization.
In personal organizations, you cannot invite members to collaborate.
If you want instead to upgrade aa team org to the scaler plan, you can create an organization with turso org create.
Do you want to continue? [y/n]: 
```

I suggest you consider changing it to something like:

```
You're upgrading your personal organization.
Plans are set and billed by organization, and your personal organization isn't your only one.
To see your organizations, use turso org list
To upgrade another organization, such as $ORG_EXAMPLE_HERE, use turso org switch $ORG_EXAMPLE_HERE

Want to update your personal org? [y/n]: y

You need to add a payment method before you can upgrade to the scaler plan.
For information about Turso plans pricing and features, access: https://turso.tech/pricing

Want to add a payment method right now? [y/n]: y

Opening your browser at:
https://billing.turso.io/p/session/test_YWNjdF8xS1U2cEFMWnl2c0E1RUlqLF9PYVVvdndTZEpxY1lOMHdPS2xoamRZbEloZ21KSWRO0100UpTZYMz9

Payment method added successfully.
You can manage your payment methods with turso org billing.

You're upgrading to the scaler plan.

Do you want to continue? [y/n]: 
```

Things important to consider:
- Prompt for confirmation before letting the user add the credit card. Adding payment information may be a tiresome process and we don't want them to realize they have to do it all over again in another org because we didn't warn them first.
- I think we should only show this warning if the user has more than one organization. If they just want to upgrade their personal org and see a message like this it may hit them the wrong way. And when we frame it like this, it doesn't make much sense to ever update a personal org - we should even consider allowing paying personal orgs to have team funcionality.

PR #620, opened 2023-09-11
Get turso db locations (add|remove) parameters from CLI arguments

athoscouto:
Examples:

```sh
> turso group list
NAME     LOCATIONS     
foo      gru (primary)     
> turso group create bar 
Created group bar at gru in 8 seconds.
> turso group locations list bar
gru (primary)
> turso group locations add bar gig
Group bar replicated to gig in 4 seconds.
> turso group locations add bar iad scl
Group bar replicated to 2 locations in 12 seconds.
> turso group locations add bar gig gru
Group bar replicated to 2 locations in 0 seconds.
```

The last example hints to the user that the replications were successful, where they are actually no-ops.
It should be fixed in the next PRs.

### DIFF
--- a/internal/cmd/group_locations.go
+++ b/internal/cmd/group_locations.go
@@ -7,6 +7,7 @@ import (
 	"github.com/chiselstrike/iku-turso-cli/internal"
 	"github.com/chiselstrike/iku-turso-cli/internal/prompt"
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/maps"
 )
 
 var groupLocationsCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


gris (approved):

internal/cmd/group_locations.go
```diff
@@ -155,3 +156,16 @@ var groupsLocationsRmCmd = &cobra.Command{
...
+func locationsCmdsArgs(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	client, err := createTursoClientFromAccessToken(false)
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveNoSpace
+	}
+	if len(args) == 0 {
+		// TODO: add completion for group names
+		return nil, cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveNoSpace
+	}
+	locations, _ := locations(client)
```
haaawk, line 169: Nit: We should probably filter out locations already present in args.

haaawk (approved):

haaawk:
Probably fixes https://github.com/tursodatabase/turso-cli/issues/616

internal/cmd/group_locations.go
athoscouto: Agree! Also, filter out locations already present in the DB itself.
But I see those as general improvements to our location autocompletion.
Let's address it separately.

PR #624, opened 2023-09-11
Group autocomplete

athoscouto:
Adds autocomplete for group commands.
- Adds autocomplete for group names (first argument of most group commands).
- Adds autocomplete for group flag in `turso db create`.
- Adds autocomplete for locations on `turso group locations add $GROUP ...`, only autocompleting missing locations.
- Adds autocomplete for locations on `turso group locations remove $GROUP ...`, only autocompleting existing locations.

### DIFF
--- a/internal/cmd/group.go
+++ b/internal/cmd/group.go
@@ -84,6 +84,7 @@ var groupsCreateCmd = &cobra.Command{
 		spinner.Stop()
 		elapsed := time.Since(start)
 		fmt.Printf("Created group %s at %s in %d seconds.\n", internal.Emph(name), internal.Emph(location), int(elapsed.Seconds()))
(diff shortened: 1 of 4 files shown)


internal/cmd/group.go
```diff
@@ -144,3 +146,41 @@ func groupsTable(groups []turso.Group) [][]string {
...
+}
+
+func getGroup(client *turso.Client, name string) (turso.Group, error) {
+	groups, err := getGroups(client)
+	if err != nil {
+		return turso.Group{}, err
+	}
+	for _, group := range groups {
+		if group.Name == name {
+			return group, nil
```
gris, line 182: just a nit: the happy path being nested made this code a bit more difficult to read. but no need to change this at all

PR #637, opened 2023-09-19
Point in time recovery

Horusiath:
This PR introduces a new command:

```
turso db restore $DATABASE --timestamp 2023-09-25T14:53:00 --yes
```

Which is going to trigger database point-in-time recovery procedure. It's going to call turso API: `/v1/databases/:database/restore`.

### DIFF
--- a/internal/cmd/db_restore.go
+++ b/internal/cmd/db_restore.go
@@ -1,0 +1,53 @@
+package cmd
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/chiselstrike/iku-turso-cli/internal"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	dbCmd.AddCommand(restoreCmd)
+	restoreCmd.Flags().BoolVarP(&yesFlag, "yes", "y", false, "Confirms the restoration of all locations of the database.")
+	addDbTimestampFlag(restoreCmd)
+	restoreCmd.RegisterFlagCompletionFunc("instance", completeInstanceName)
+}
+
+var restoreCmd = &cobra.Command{
+	Use:               "restore database_name --timestamp 2023-09-25T14:53:00",
+	Short:             "Restore a database to a given point in time.",
+	Args:              cobra.ExactArgs(1),
+	ValidArgsFunction: dbNameArg,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		client, err := createTursoClientFromAccessToken(true)
+		if err != nil {
+			return err
+		}
+		name := args[0]
+		timestamp, err := time.Parse("2006-01-02T15:04:05", timestampFlag)
+		if err != nil {
+			return fmt.Errorf("could not parse timestamp using 'yyyy-MM-ddThh:mm:ss' pattern: %w", err)
+		}
+
+		if yesFlag {
+			return restoreDatabase(client, name, timestamp)
+		}
+
+		fmt.Printf("Database %s, all its replicas, and data will be replaced.\n", internal.Emph(name))
+
+		ok, err := promptConfirmation("Are you sure you want to do this?")
+		if err != nil {
+			return fmt.Errorf("could not get prompt confirmed by user: %w", err)
+		}
+
+		if !ok {
+			fmt.Println("Database restoration cancelled.")
+			return nil
+		}
+
+		return restoreDatabase(client, name, timestamp)
+	},
+}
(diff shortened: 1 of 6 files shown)


internal/turso/turso.go
```diff
@@ -166,3 +166,7 @@ func (t *Client) Upload(path string, fileData *os.File) (*http.Response, error)
 func (t *Client) Delete(path string, body io.Reader) (*http.Response, error) {
 	return t.do("DELETE", path, body)
 }
+
+func (t *Client) Restore(path string, body io.Reader) (*http.Response, error) {
+	return t.do("DELETE", path, body)
```
haaawk, line 171: Is restore a HTTP DELETE?

internal/cmd/timestamp_flag.go
```diff
@@ -0,0 +1,9 @@
+package cmd
+
+import "github.com/spf13/cobra"
+
+var timestampFlag string
+
+func addDbTimestampFlag(cmd *cobra.Command) {
+	restoreCmd.Flags().StringVarP(&timestampFlag, "timestamp", "t", "", "UTC timestamp for the version of database at a given point in time.")
```
haaawk, line 8: What's the format of the timestamp? Just int? In what unit?

haaawk: Ok. I can see it here https://github.com/tursodatabase/turso-cli/pull/637/files#diff-e963bae42bbc05e1ecbaa07a34c3a585f0278493311e24f2a4ab030e8267c940R19

internal/cmd/db_restore.go
```diff
@@ -0,0 +1,53 @@
...
+
+func init() {
+	dbCmd.AddCommand(restoreCmd)
+	restoreCmd.Flags().BoolVarP(&yesFlag, "yes", "y", false, "Confirms the restoration of all locations of the database.")
+	addDbTimestampFlag(restoreCmd)
+	restoreCmd.RegisterFlagCompletionFunc("instance", completeInstanceName)
+}
+
+var restoreCmd = &cobra.Command{
+	Use:               "restore database_name --timestamp 2023-09-25T14:53:00",
```
haaawk, line 19: I think timestamp should not be a flag but an argument. Without timestamp this operation makes no sense. You're also not checking if timestamp is set in the RunE. Which just proves my point.

internal/cmd/db_restore.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/chiselstrike/iku-turso-cli/internal"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	dbCmd.AddCommand(restoreCmd)
+	restoreCmd.Flags().BoolVarP(&yesFlag, "yes", "y", false, "Confirms the restoration of all locations of the database.")
+	addDbTimestampFlag(restoreCmd)
+	restoreCmd.RegisterFlagCompletionFunc("instance", completeInstanceName)
```
haaawk, line 15: This should probably go away, no?

internal/cmd/utils.go
```diff
@@ -131,6 +131,22 @@ func printTable(header []string, data [][]string) {
...
+	defer s.Stop()
+
+	if err := client.Databases.Restore(name, timestamp); err != nil {
+		return err
+	}
+	s.Stop()
+	elapsed := time.Since(start)
+
+	fmt.Printf("Restored database %s in %d seconds.\n", internal.Emph(name), int(elapsed.Seconds()))
+	invalidateDatabasesCache()
```
haaawk, line 146: We don't need this, do we?

internal/cmd/db_restore.go
Horusiath: General idea was that in the future we might want to do restoration based on different kinds of parameters (sqld natively supports timestamp-based or generation-based restoration).

internal/cmd/timestamp_flag.go
```diff
@@ -0,0 +1,9 @@
+package cmd
+
+import "github.com/spf13/cobra"
+
+var timestampFlag string
+
+func addDbTimestampFlag(cmd *cobra.Command) {
```
haaawk, line 7: Is this used anywhere?

haaawk (approved):

Horusiath:
After yesterday's talks and discussion with @athoscouto it looks like it will be more reasonable to change that feature into optional param as part of `turso db create {forked-db-name} --from-db {parent-db-name} --timestamp {timestamp}`.

PR #638, opened 2023-09-19
Simplify turso db create to only work with groups

athoscouto:
This simplifies `turso db create`, removing support for databases that do not use groups.

It was kept simple by choice, we'll add other features (e.g. first-time hint) as we go.
We need to implement:
- Support for canary images
- Support for extensions

----

This is the first of a series of PRs targeted at `athos/groups-as-default` which will make all CLI commands work only with database groups.
The plan is to use the target branch as a feature one, merge all PRs to it, and then merge it to `main`.  

----

This severely changes the behavior of the `--group` flag.
- If the flag is set, the specified group must exist (matches current behavior).
- If the flag is not set:
    - If there are no groups, the `default` group is created.
    - If there is a single group, the database is created in that group.
    - If there are two or more groups, the command errors out.

The rationale here is not to require users to understand how groups work unless they have more than one.
A valid alternative is to always default `--group` to `default`, but that would be annoying if you have a single group with another name. 

As a principle, groups should be explicitly created - to prevent mistakes. The only exception is automatically creating the default group, ONLY when there are no groups, to make starting with `turso` easier.

### DIFF
--- a/internal/cmd/db_create.go
+++ b/internal/cmd/db_create.go
@@ -40,33 +40,39 @@ var createCmd = &cobra.Command{
 			return err
 		}
 
-		locationId := locationFlag
-		if locationId == "" {
-			locationId, _ = closestLocation(client)
-		}
-		if !isValidLocation(client, locationId) {
-			return fmt.Errorf("location '%s' is not a valid one", locationId)
+		group, err := groupFromFlag(client)
+		if err != nil {
+			return err
 		}
 
-		locationText := fmt.Sprintf("%s (%s)", locationDescription(client, locationId), locationId)
+		location, err := locationFromFlag(client)
+		if err != nil {
+			return err
+		}
+
+		if ok, err := shouldCreateGroup(client, group, location); ok {
+			createGroup(client, group, location)
+		} else if err != nil {
+			return err
+		}
+
 		start := time.Now()
-		description := fmt.Sprintf("Creating database %s in %s", internal.Emph(name), internal.Emph(locationText))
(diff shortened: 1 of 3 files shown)


internal/cmd/db_create.go
```diff
@@ -132,211 +35,110 @@ var createCmd = &cobra.Command{
...
+		location, err := locationFromFlag(client)
 		if err != nil {
 			return err
 		}
 
-		dbText := ""
-		if fromFileFlag != "" {
-			dbText = fmt.Sprintf(" from file %s", internal.Emph(fromFileFlag))
+		if ok, err := shouldCreateGroup(client, group, location); ok {
+			createGroup(client, group, location)
```
avinassh, line 54: ⚠️ we are ignoring `createGroup` error 

also, how about we simplify this and avoid `else if`:

```go
	ok, err := shouldCreateGroup(client, group, location)
	if err != nil {
		return err
	}
	if ok {
		if err = createGroup(client, group, location); err != nil {
			return err
		}
	}
```

avinassh:
> The rationale here is not to require users to understand how groups work unless they have more than one.

we do mention that we are creating the db in a group. Won't be confusing for a new user who is trying to create a db  and the cmd says 'created a database in default group'?

avinassh:
> It was kept simple by choice, we'll add other features (e.g. first-time hint) as we go.

I would suggest adding the hint too. More people are going to use it next week and so hint will be helpful

avinassh (approved):

PR #641, opened 2023-09-19
Error out when --location or --instance is flag on db destroy for group dbs

athoscouto:
See https://github.com/tursodatabase/turso-cli/pull/638.

### DIFF
--- a/internal/cmd/db_destroy.go
+++ b/internal/cmd/db_destroy.go
@@ -26,12 +26,24 @@ var destroyCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
+
+		db, err := getDatabase(client, args[0])
+		if err != nil {
+			return nil
+		}
+
 		name := args[0]
 		if instanceFlag != "" {
+			if db.Group != "" {
+				return fmt.Errorf("group databases do not support instance destruction.\nUse %s instead", internal.Emph("turso group locations rm "+name))
+			}
 			return destroyDatabaseInstance(client, name, instanceFlag)
 		}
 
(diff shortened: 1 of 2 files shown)


internal/cmd/db_destroy.go
```diff
@@ -26,20 +26,32 @@ var destroyCmd = &cobra.Command{
...
 		}
+
+		db, err := getDatabase(client, args[0])
+		if err != nil {
+			return nil
+		}
+
 		name := args[0]
 		if instanceFlag != "" {
+			if db.Group != "" {
```
avinassh, line 37: we are not using the same logic which we use in replicate https://github.com/tursodatabase/turso-cli/pull/642#discussion_r1333977267

i.e. for people who have only one group, we will do the `locations rm`

avinassh (approved):

internal/cmd/db_destroy.go
avinassh: also this:

> The rationale here is not to require users to understand how groups work unless they have more than one.
A valid alternative is to always default --group to default, but that would be annoying if you have a single group with another name.

athoscouto: That is intended. Replicate is on most of our tutorials and is one of the main commands of the CLI. `turso db destroy --instance` and `turso db destroy --location` always felt like a stretch of `destroy` because `replicate` had no revert op. With `locations add/rm` now we do have a set of cmds which is more sound. Let's nudge people towards using that.
One thing we can do later is add `turso db locations add/rm` for a single DB.

PR #642, opened 2023-09-19
Make turso db replicate work with groups

athoscouto:
See https://github.com/tursodatabase/turso-cli/pull/638.

### DIFF
--- a/internal/cmd/db_replicate.go
+++ b/internal/cmd/db_replicate.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/chiselstrike/iku-turso-cli/internal"
 	"github.com/chiselstrike/iku-turso-cli/internal/prompt"
-	"github.com/chiselstrike/iku-turso-cli/internal/settings"
 	"github.com/chiselstrike/iku-turso-cli/internal/turso"
 	"github.com/manifoldco/promptui"
 	"github.com/rodaine/table"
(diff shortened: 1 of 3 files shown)


internal/cmd/db_replicate.go
```diff
@@ -186,17 +126,103 @@ func handleDatabaseReplicationError(client *turso.Client, name, locationId strin
...
+		counter[database.Group]++
+		if database.Name == name {
+			group = database.Group
+		}
+	}
+
+	if group == "" {
+		return true, nil
+	}
+	return counter[group] == 1, nil
```
avinassh, line 227: I'd like to understand the logic: 

1. If the db doesn't belong to any group, we will let `db replicate` work (existing behaviour)
2. If there is only one group, then we let `db replicate` work, I guess this is assuming everyone would have a `default` group and we don't want users to annoy
3. If there are more than one group, then we ask them use `group locations add`

avinassh (approved):

internal/cmd/db_replicate.go
athoscouto: > 2. If there is only one ~group~ database in the group, then we let db replicate work, ~I guess this is assuming everyone would have a default group and we don't want users to annoy~. Because if the group has only a single DB, it works like there are no groups!
> 3. If there are more than one ~group~ database in the group, then we ask them use group locations add

PR #643, opened 2023-09-20
Enabled timestamp param in turso db create

Horusiath:
Related to https://github.com/libsql/sqld/pull/690

Example scenario: 

```bash
turso group create {group}
turso db create {source} --group {group}

# do some work on {source} then fork it with state version at given point in time
turso db create {dest} --from-db={source} --group={group} --timestamp=2023-09-20T14:30:00
```

### DIFF
--- a/internal/cmd/db_create.go
+++ b/internal/cmd/db_create.go
@@ -165,6 +165,17 @@ var createCmd = &cobra.Command{
 			extensions = "all"
 		}
 
+		var timestamp *time.Time
+		if fromDBFlag != "" {
+			ts, err := time.Parse("2006-01-02T15:04:05", timestampFlag)
+			if err != nil {
+				return fmt.Errorf("provided timestamp was not in 'yyyy-MM-ddThh:mm::ss' format")
+			}
+			timestamp = &ts
+		} else if timestampFlag != "" {
+			return fmt.Errorf("--timestamp cannot be used without specifying --from-db option")
+		}
+
 		dbFile, err := getDbFile(fromFileFlag)
 		if err != nil {
 			return err
@@ -180,7 +191,7 @@ var createCmd = &cobra.Command{
 		description := fmt.Sprintf("Creating database %s%s in %s", internal.Emph(name), dbText, internal.Emph(locationText))
 		spinner := prompt.Spinner(description)
 		defer spinner.Stop()
-		if _, err = client.Databases.Create(name, locationId, image, extensions, groupFlag, fromDBFlag); err != nil {
+		if _, err = client.Databases.Create(name, locationId, image, extensions, groupFlag, fromDBFlag, timestamp); err != nil {
 			return fmt.Errorf("could not create database %s: %w", name, err)
 		}
 
--- a/internal/cmd/group_flag.go
+++ b/internal/cmd/group_flag.go
@@ -18,9 +18,13 @@ func addGroupFlag(cmd *cobra.Command) {
 }
 
 var fromDBFlag string
+var timestampFlag string
 
 func addFromDBFlag(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&fromDBFlag, "from-db", "", "Creates the new database based on an existing one")
 	cmd.Flags().MarkHidden("from-db")
 	cmd.RegisterFlagCompletionFunc("from-db", dbNameArg)
+
+	cmd.Flags().StringVar(&timestampFlag, "timestamp", "", "When used with --from-db option, fork will represent state of the origin database at given point in time")
+	cmd.Flags().MarkHidden("timestamp")
 }
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@ completions
 internal/cmd/version.txt
 .turso_history
 .vscode/
+.idea/
 cmd/turso/turso
--- a/internal/turso/databases.go
+++ b/internal/turso/databases.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"time"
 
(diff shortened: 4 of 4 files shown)


internal/cmd/db_create.go
```diff
@@ -165,6 +165,17 @@ var createCmd = &cobra.Command{
 			extensions = "all"
 		}
 
+		var timestamp *time.Time
+		if fromDBFlag != "" {
+			ts, err := time.Parse("2006-01-02T15:04:05", timestampFlag)
```
haaawk, line 170: Is timestamp flag required for forking? I thought it's also possible to just for current db, no?

haaawk (changes requested):
Looks good to me except that using fromDBFlag without timestampFlag won't work any more and I don't think it's what we want.

internal/cmd/group_flag.go
```diff
@@ -18,9 +18,13 @@ func addGroupFlag(cmd *cobra.Command) {
...
 
 var fromDBFlag string
+var timestampFlag string
 
 func addFromDBFlag(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&fromDBFlag, "from-db", "", "Creates the new database based on an existing one")
 	cmd.Flags().MarkHidden("from-db")
 	cmd.RegisterFlagCompletionFunc("from-db", dbNameArg)
+
+	cmd.Flags().StringVar(&timestampFlag, "timestamp", "", "When used with --from-db option, fork will represent state of the origin database at given point in time")
```
athoscouto, line 28: ⛏️: please don't mention "fork" for now, since we're not using it anywhere else in the product. Some other suggestions:
```suggestion
	cmd.Flags().StringVar(&timestampFlag, "timestamp", "", "Must be used with --from-db flag. The created database will contain the data of the original one at the point in time specified by the timestamp flag")
```

internal/cmd/db_create.go
```diff
@@ -165,6 +165,17 @@ var createCmd = &cobra.Command{
...
+			if err != nil {
+				return fmt.Errorf("provided timestamp was not in 'yyyy-MM-ddThh:mm::ss' format")
+			}
+			timestamp = &ts
+		} else if timestampFlag != "" {
+			return fmt.Errorf("--timestamp cannot be used without specifying --from-db option")
+		}
```
athoscouto, line 177: ⛏️: please extract this into a function that returns (*time.Time, error)`. You can use early returns to avoid nesting conditionals there. 

As @haaawk pointed out, please also handle the `timestampFlag == ""` case.

internal/cmd/group_flag.go
```diff
@@ -18,9 +22,29 @@ func addGroupFlag(cmd *cobra.Command) {
...
+			if err != nil {
+				return nil, fmt.Errorf("provided timestamp was not in 'yyyy-MM-ddThh:mm::ss' format")
+			}
+			timestamp = &ts
+		}
+	} else if timestampFlag != "" {
+		return nil, fmt.Errorf("--timestamp cannot be used without specifying --from-db option")
+	}
+	return timestamp, nil
 }
```
athoscouto, line 50: ⛏️: I suggest the following refactoring:
```suggestion
func parseTimestampFlag() (*time.Time, error) {
	if timestampFlag == "" {
		return nil, nil
	}
	if fromDBFlag == "" {
		return nil, fmt.Errorf("--timestamp cannot be used without specifying --from-db")
	}

	ts, err := time.Parse("2006-01-02T15:04:05", timestampFlag)
	if err != nil {
		return nil, fmt.Errorf("provided timestamp was not in 'yyyy-MM-ddThh:mm::ss' format")
	}
	return ts, nil
}
```
With it, you'll explicitly handle the cases you want to handle while avoiding nesting conditional logic.

internal/cmd/db_create.go
```diff
@@ -165,6 +165,17 @@ var createCmd = &cobra.Command{
...
 
+		var timestamp *time.Time
+		if fromDBFlag != "" {
+			ts, err := time.Parse("2006-01-02T15:04:05", timestampFlag)
+			if err != nil {
+				return fmt.Errorf("provided timestamp was not in 'yyyy-MM-ddThh:mm::ss' format")
+			}
+			timestamp = &ts
+		} else if timestampFlag != "" {
+			return fmt.Errorf("--timestamp cannot be used without specifying --from-db option")
```
haaawk: Why is this always an error when timestampFlag is ""?

haaawk (changes requested):
LGTM except the missing --timestamp flag is still causing a failure

athoscouto (approved):
LGTM

I have a nit regarding the structure of the `parseTimestampFlag` function.
But your version should be functionally equivalent to mine, so we're good to go.

internal/cmd/db_create.go
```diff
@@ -165,6 +165,11 @@ var createCmd = &cobra.Command{
 			extensions = "all"
 		}
 
+		timestamp, err := parseTimestampFlag()
+		if err != nil {
+			return nil
```
athoscouto, line 170:
```suggestion
			return err
```

internal/cmd/group_flag.go
athoscouto: @Horusiath as @haaawk they are not functionally equivalent. Yours is missing the case where both flags are empty.
You should also address my other suggestion, which is preventing errors from being displayed to the user.

internal/cmd/db_create.go
Horusiath: It's an error when the `--timestamp` was defined while `--from-db` was not. Timestamp has only sense when used together with --from-db (required by namespace forking).

haaawk: got it. I confused != "" with == ""

haaawk (approved):

PR #647, opened 2023-09-22
fix: add logic to put credit card if user doesn't have yet

gris:


internal/cmd/plan.go
```diff
@@ -202,6 +202,21 @@ var planEnableOverages = &cobra.Command{
...
+		if err != nil {
+			return err
+		}
+		if !hasPaymentMethod {
+			ok, err := PaymentMethodHelperOverages(client)
+			if err != nil {
+				return fmt.Errorf("failed to check payment method: %w", err)
+			}
+			if !ok {
+				return nil
```
athoscouto, line 215: You should either return an error, or print a feedback saying that overages weren't enabled.

internal/cmd/plan.go
```diff
@@ -315,6 +330,44 @@ func PaymentMethodHelper(client *turso.Client, selected string) (bool, error) {
 	}
 }
 
+func PaymentMethodHelperOverages(client *turso.Client) (bool, error) {
```
athoscouto, line 333: We have 3 almost identical functions like this one. Anything we can do to make this logic more concise?

athoscouto (approved):

internal/cmd/plan.go
```diff
@@ -311,9 +311,19 @@ func PaymentMethodHelper(client *turso.Client, selected string) (bool, error) {
...
 	errsInARoW := 0
+	var hasPaymentMethod bool
+	var err error
 	for {
-		hasPaymentMethod, err := client.Billing.HasPaymentMethod()
+		if stripeId != "" {
+			hasPaymentMethod, err = client.Billing.HasPaymentMethodWithStripeId(stripeId)
+		} else {
+			hasPaymentMethod, err = client.Billing.HasPaymentMethod()
+		}
```
athoscouto, line 326: To avoid this, you can extract the conditional into a function:

```go
func hasPaymentMethod(client *turso.Client, stripeId string) (bool, error) {
	if stripeId != "" {
		return client.Billing.HasPaymentMethodWithStripeId(stripeId)
	}
	return client.Billing.HasPaymentMethod()
}
```

PR #668, opened 2023-09-29
feat: add autoupdate capability

gris:
- New command to manage autoupdates
- `turso autoupdate enable` and `turso autoupdate disable`
- In case `autoupdate` is enabled it checks once a day if there is a new latest available, in case there is, update it.

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/chiselstrike/iku-turso-cli/internal/flags"
 	"github.com/chiselstrike/iku-turso-cli/internal/settings"
@@ -36,7 +37,26 @@ func init() {
 		settings.PersistChanges()
 	}
 	rootCmd.CompletionOptions.HiddenDefaultCmd = true
-
+	rootCmd.PersistentPostRun = func(cmd *cobra.Command, args []string) {
+		settings, err := settings.ReadSettings()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to read settings: %s\n", err)
+			os.Exit(1)
+		}
+		if settings.GetAutoupdate() && time.Now().Unix() > settings.GetLastUpdateCheck()+int64(24*60*60) {
+			fmt.Println("Checking for updates...")
+			latest, err := fetchLatestVersion()
+			settings.SetLastUpdateCheck(time.Now().Unix())
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "failed to get latest version: %s\n", err)
+				return
+			}
+			if version != "dev" && version < latest {
+				Update()
+				return
+			}
+		}
+	}
 	flags.AddDebugFlag(rootCmd)
 	flags.AddResetConfigFlag(rootCmd)
 }
--- a/internal/cmd/update.go
+++ b/internal/cmd/update.go
@@ -6,14 +6,19 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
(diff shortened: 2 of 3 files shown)


internal/cmd/root.go
```diff
@@ -36,7 +37,26 @@ func init() {
 		settings.PersistChanges()
 	}
 	rootCmd.CompletionOptions.HiddenDefaultCmd = true
-
+	rootCmd.PersistentPostRun = func(cmd *cobra.Command, args []string) {
```
haaawk, line 40: Aren't you you overriding the rootCmd.PersistentPostRun set above in line 35/36 that calls `settings.PersistentChanges()`

internal/cmd/update.go
```diff
@@ -6,14 +6,19 @@ import (
...
 
 	"github.com/chiselstrike/iku-turso-cli/internal"
+	"github.com/chiselstrike/iku-turso-cli/internal/settings"
 )
 
 func init() {
 	rootCmd.AddCommand(updateCmd)
+	rootCmd.AddCommand(autoupdateCmd)
+	autoupdateCmd.AddCommand(autoupdateEnableCmd)
+	autoupdateCmd.AddCommand(autoupdateDisableCmd)
```
haaawk, line 21: Why are you having two commands for autoupdate?

internal/cmd/update.go
```diff
@@ -36,14 +41,53 @@ func IsUnderHomebrew() bool {
...
+	Args:              cobra.NoArgs,
+	ValidArgsFunction: noFilesArg,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		settings, err := settings.ReadSettings()
+		if err != nil {
+			return fmt.Errorf("failed to read settings: %w", err)
+		}
+		settings.SetAutoupdate(true)
+		settings.SetLastUpdateCheck(time.Now().Unix())
```
haaawk, line 61: I would set it to 0 so that we actually do a version check once someone enables it.

haaawk (commented):
Looks mostly good.

internal/cmd/root.go
```diff
@@ -34,9 +36,34 @@ func init() {
 
 	rootCmd.PersistentPostRun = func(cmd *cobra.Command, args []string) {
 		settings.PersistChanges()
+		if version == "dev" {
+			return
+		}
+		settings, err := settings.ReadSettings()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to read settings: %s\n", err)
```
athoscouto, line 44: This is a background op that is not related to what the user wanted to do.
WDYT about failing silently?

internal/cmd/root.go
```diff
@@ -34,9 +36,34 @@ func init() {
...
+			os.Exit(1)
+		}
+		if settings.GetAutoupdate() == "on" && time.Now().Unix() >= settings.GetLastUpdateCheck()+int64(24*60*60) {
+			fmt.Println("Checking for updates...")
```
athoscouto, line 48: Checking for updates should be pretty quick.
I'd make it silent unless we actually trigger the update.

internal/settings/settings.go
```diff
@@ -119,3 +120,44 @@ func (s *Settings) SetUsername(username string) {
...
+	}
+	return lastUpdateCheck.(int64)
+}
+
+func (s *Settings) GetAutoupdate() string {
+	config := viper.GetStringMap("config")
+	if config == nil {
+		return "on"
+	}
+	value := config["autoupdate"]
```
athoscouto, line 161: The value itself could also be empty. We should probably return `on` when that happens as well.

athoscouto (approved):

PR #669, opened 2023-09-29
Support deleting multiple databases with turso db destroy

phenpessoa:
First time contributing to the repo, sorry if I'm doing something wrong, couldn't find a contributions guideline.

fixes https://github.com/tursodatabase/turso-cli/issues/667

### DIFF
--- a/internal/cmd/db_destroy.go
+++ b/internal/cmd/db_destroy.go
@@ -65,7 +65,7 @@ func handleDestroySingleDB(args []string, client *turso.Client) error {
 		return destroyDatabases(client, args)
 	}
 
-	fmt.Printf("Database %s, and all its data will be destroyed.\n", internal.Emph(name))
+	fmt.Printf("Database %s and all its data will be destroyed.\n", internal.Emph(name))
 
 	ok, err := promptConfirmation("Are you sure you want to do this?")
 	if err != nil {
--- a/internal/cmd/db_destroy.go
+++ b/internal/cmd/db_destroy.go
@@ -32,10 +32,10 @@ var destroyCmd = &cobra.Command{
 		}
 
 		if len(args) > 1 {
-			return handleDestroySingleDB(args, client)
+			return handleDestroyMultipleDBs(args, client)
 		}
 
-		return handleDestroyMultipleDBs(args, client)
+		return handleDestroySingleDB(args, client)
 	},
 }
 
--- a/internal/cmd/db_destroy.go
+++ b/internal/cmd/db_destroy.go
@@ -61,7 +61,7 @@ var destroyCmd = &cobra.Command{
 			return destroyDatabases(client, args)
 		}
 
-		if len(args) > 0 {
+		if len(args) > 1 {
 			fmt.Printf("Databases %s and all their data will be destroyed.\n", internal.Emph(strings.Join(args, ", ")))
 		} else {
 			fmt.Printf("Database %s, and all its data will be destroyed.\n", internal.Emph(args[0]))
--- a/internal/cmd/db_destroy.go
+++ b/internal/cmd/db_destroy.go
@@ -1,10 +1,12 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/chiselstrike/iku-turso-cli/internal"
+	"github.com/chiselstrike/iku-turso-cli/internal/turso"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


penberg:
Thanks @phenpessoa! @athoscouto @gris @avinassh, can you folks please review?

internal/cmd/utils.go
```diff
@@ -131,23 +131,34 @@ func printTable(header []string, data [][]string) {
...
-	s := prompt.Spinner(fmt.Sprintf("Destroying database %s... ", internal.Emph(name)))
+	destroy := func(name string) error {
+		start := time.Now()
+
+		if err := client.Databases.Delete(name); err != nil {
+			return err
+		}
+		elapsed := time.Since(start)
+
+		fmt.Printf("Destroyed database %s in %d seconds.\n", internal.Emph(name), int(elapsed.Seconds()))
```
athoscouto, line 147: When this print is executed the spinner is still running with `Destroying databases...`.
This is a problem because all widgets from the `prompt` package expect that they are the only ones writing to `stdout` while they are active.
To make matters worse, when this assumption is broken the CLI gets to a weird state where it doesn't respond well - to fix it you need to either nuke the shell or reset it.

To keep things simple, my suggestion is to remove any printing logic from `destroy`, time the execution of the whole `errgroup`, and:
- If you deleted a single database, print: `fmt.Printf("Destroyed database %s in %d seconds.\n", internal.Emph(names[0]), int(elapsed.Seconds()))`
- if you deleted more than one database, print: `fmt.Printf("Destroyed %d databases in %d seconds.\n", len(names), int(elapsed.Seconds()))`

athoscouto: Or do as you did on the are you sure prompt, joining their names. Be aware that the output can get weird if you are deleting too many databases.

internal/cmd/db_destroy.go
```diff
@@ -27,31 +28,44 @@ var destroyCmd = &cobra.Command{
...
-			return destroyDatabaseRegion(client, name, locationFlag)
 		}
 
 		if yesFlag {
-			return destroyDatabase(client, name)
+			return destroyDatabases(client, args)
 		}
 
-		fmt.Printf("Database %s, and all its data will be destroyed.\n", internal.Emph(name))
+		if len(args) > 0 {
```
athoscouto, line 64:
```suggestion
		if len(args) > 1 {
```

internal/cmd/db_destroy.go
```diff
@@ -27,31 +28,44 @@ var destroyCmd = &cobra.Command{
 			return err
 		}
 
-		db, err := getDatabase(client, args[0])
-		if err != nil {
-			return nil
-		}
+		if instanceFlag != "" || locationFlag != "" {
```
athoscouto, line 31: These flags are only supposed to be used on databases that were created before database groups existed. In those databases, instance names are unique - so if you're specifying an instance you shouldn't use more than one DB name.
So if either flags are set I think you should error out if `len(args) > 1`.

athoscouto (changes requested):
Thanks for the contribution!

Made some comments. The spinner one is the most critical, so we'll need that fixed before moving forward. Let me know if you have any questions.

internal/cmd/db_destroy.go
phenpessoa: Thank you! I wasn't sure what to do on this case.
Will make the changes and push.

phenpessoa:
@athoscouto pushed the requested changes

internal/cmd/db_destroy.go
```diff
@@ -18,51 +21,85 @@ func init() {
...
+	}
+
+	return destroyDatabases(client, args)
+}
+
+func handleDestroyMultipleDBs(args []string, client *turso.Client) error {
+	if instanceFlag != "" || locationFlag != "" {
+		return errors.New("can not use location nor instance flag when deleting more than 1 database")
+	}
+
```
athoscouto, line 87: What happens when you try to destroy a DB that doesn't exist? Is the error message clear?
When destroying a single one we do `getDatabase`, so we exit early if the DB does not exist.

phenpessoa: Is this question related to when the instance or the location flag is passed or in general?

I ran `turso db destroy idk idk2 idk3`, where idk3 doesnt exist, and got `Error: database idk3 not found. List known databases using turso db list`, and idk1 and 2 got deleted, but this is "random", it was just a coincidence that idk3's  closure ended lastly.

If I ran it with only dbs that doesn't exit, one of the listed dbs will be used as an error "randomly".

I'm not sure how to handle each error individually using a waitgroup, because it always stop at the first error encountered. From what I understand we have 2 options:
1 - run the code linearly, and stop at the first error, printing the success msgs that happened before this error occurred, if any
2 - create a "custom concurrent error handler" and print all errors/success messages to the console

I think 2 would make more sense, I'm happy to work on it. What do you think?

athoscouto: I think what you have right now is good enough!

PR #678, opened 2023-10-05
feat: adjust signup URL params

viniciuslacerda0:
Fixes [FRO-353](https://linear.app/chiselstrike/issue/FRO-353/add-to-turso-cli-signup-with-type-=cli-signup)

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -222,6 +222,7 @@ func beginAuth(port int, headless bool, path string) (string, error) {
 		authUrl.RawQuery = url.Values{
 			"port":     {strconv.Itoa(port)},
 			"redirect": {"true"},
+			"type":     {"cli"},
 		}.Encode()
 		err = browser.OpenURL(authUrl.String())
 		if err != nil {


internal/cmd/auth.go
```diff
@@ -213,16 +215,22 @@ func auth(cmd *cobra.Command, args []string, path string) error {
...
+func beginAuth(port int, headless bool, path string, extraQueryParams map[string]string) (string, error) {
 	authUrl, err := url.Parse(fmt.Sprintf("%s%s", getTursoUrl(), path))
 	if err != nil {
 		return "", fmt.Errorf("error parsing auth URL: %w", err)
 	}
 	if !headless {
-		authUrl.RawQuery = url.Values{
+		rawQuery := url.Values{
 			"port":     {strconv.Itoa(port)},
 			"redirect": {"true"},
```
athoscouto, line 226: Are there different signup flows in the CLI?
If not, just add the `type` directly here instead of passing it all the way from `signup`.

viniciuslacerda0: If I don't do this, when a user tries to login, the `type=cli-signup` will be added too, and we don't want that to happen

viniciuslacerda0: I could change the `extraQueryParams` to be a string and add it to the `url.Values` if is not `nil`, but I think that having this map would allow us to send more queryParams if needed in the future

WilsonNet (approved):
I think this is an elegant solution with the map for parameters

internal/cmd/auth.go
athoscouto: But aren't they using different paths? You could always add `type=cli` and differ sign ups from sign ins by the path, no?

viniciuslacerda0: I tried changing only in the path (line 121, e.g. `return auth(cmd, args, "/signup?type=cli-signup")`) and it didn't work (since the authUrl.RawQuery is rewritten), In the commit a7b1c134085e68f8858773e3648f13ac7a3e2070 I tried another approach verifying the path to add this type to the query.

athoscouto (approved):

PR #680, opened 2023-10-06
Allow creating new DBs from CSV files

### LINKED ISSUE #604: Add a way to import CSV files
Right now there is no easy way to import a CSV file. A work around is: 

1. create a local sqlite DB with the CSV file: 
    ```sh
    $ sqlite data.db
    
    > .mode csv
    > PRAGMA journal_mode=WAL;
    > .import /path-to/file/users.csv users
    ```
2. Use that to seed new database:
    ```
    $ turso db create --from-file data.db
    ```

This is not great. So we should add a way to import CSV file into a new or existing database.  The commands could be: 

```
turso db create --from-csv-file users.csv --csv-table-name users
```

For existing databases, we can add an option in shell - issue: https://github.com/libsql/libsql-shell-go/issues/152

avinassh:
closes #604

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.21
 toolchain go1.21.0
 
 require (
+	github.com/Clever/csvlint v0.3.0
 	github.com/athoscouto/codename v0.0.3
 	github.com/charmbracelet/bubbles v0.15.0
 	github.com/charmbracelet/bubbletea v0.23.1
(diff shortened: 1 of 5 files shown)


gris (approved):

athoscouto:
@avinassh how can we allow users to use the same feature on existing DBs?

athoscouto (approved):

avinassh:
@athoscouto we don't support `.import` statements in the turso shell. So we need to do a similar work around, generate a SQLite db, take the export and then import them using `<` in shell:

```
$ turso db shell funky-falcon < my-csv-table.sql
```

I am tracking here #712

avinassh:
I just realised that we don't import and create a table anywhere. So I will probably add a new sub command `db import`

PR #686, opened 2023-10-10
Fix CI cache problems

athoscouto:


README.md
```diff
@@ -1,15 +1,17 @@
 # Turso CLI
 
-[![goreleaser](https://github.com/tursodatabase/turso-cli/actions/workflows/goreleaser.yml/badge.svg)](https://github.com/tursodatabase/turso-cli/actions/workflows/goreleaser.yml)
+[![Go](https://github.com/tursodatabase/turso-cli/actions/workflows/go.yml/badge.svg)](https://github.com/tursodatabase/turso-cli/actions/workflows/go.yml)
```
haaawk, line 3: Why are you removing goreleaser? I think both are useful.

haaawk (approved):
LGTM. Is cache really useless in self hosted runners? Is it because we don't clean up the runners between CI runs?

README.md
athoscouto: Originally we just used `go.yml`. 

But [it always failed ](https://github.com/tursodatabase/turso-cli/actions/runs/6461516569) because of conflicts when we restored the cache.
Screenshot 2023-10-10 at 8 26 05 AM
This resulted in the errors you see above and also on a red status badge.
That is why @gris changed it to `goreleaser.yml` a few days ago.

I think a CI badge makes more sense. Meaning that the "last CI run was successful for this project".
I don't see much value in a badge saying "the last release was successful". But I also don't oppose having it.
Let me know if you still want to keep `gorereleaser` badge and I'll add it back. @gris let me know what you think as well.

athoscouto:
@haaawk I think so. I was reading issues on setup-go repo to understand what was the problem here and came across [this](https://github.com/actions/setup-go/issues/403#issuecomment-1659607846).
It is also a bit harmful since we transfer that huge amount of data (300MB+) on every run and it resulted in errors that made our badge look broken.

gris (approved):

README.md
gris: Makes a lot of sense! Thanks for spotting that

PR #692, opened 2023-10-16
Add wait flag to group creation and location add

athoscouto:


internal/turso/groups.go
```diff
@@ -153,6 +153,24 @@ func (d *GroupsClient) RemoveLocation(name, location string) error {
 	return nil
 }
 
+func (d *GroupsClient) WaitLocation(name, location string) error {
+	res, err := d.client.Get(d.URL("/"+name+"/locations/"+location+"/wait"), nil)
```
gris, line 157: didn't know that this would work. Nice

gris:
Should we always wait by default?

athoscouto:
I don't think so. The platform should shield users from this.
It is only important in test settings when we want to make sure a group or replica is up before moving to the next step.

PR #695, opened 2023-10-17
Update libsql docs URL

JakeHedman:
Link died when sqld moved to tursodatabase/libsql


gris (approved):
Thanks for the contribution

Issue #701, opened 2023-10-24
Improve logged out message


penberg:
When I try to log in, but the token is expired I get this message:

```
penberg@vonneumann ~ % turso db create demo
Error: user not logged in, please login with turso auth login
```

Technically, this is _not_ an error, but really part of normal life. Let's change the message to something like:

```
penberg@vonneumann ~ % turso db create demo
You are not logged in, please login with `turso auth login` before running other commands.
```

PR #703, opened 2023-10-24
Solved #701: Added improved logged out message

PruthviThakor:
- changed logged out message to **You are not logged in, please login with `turso auth login` before running other commands.** as mentioned in issues #701 .

### DIFF
--- a/internal/cmd/db.go
+++ b/internal/cmd/db.go
@@ -112,7 +112,7 @@ func getAccessToken(warnMultipleAccessTokenSources bool) (string, error) {
 		return envToken, nil
 	}
 	if !isJwtTokenValid(settingsToken) {
-		return "", fmt.Printf("You are not logged in, please login with %s before running other commands.", internal.Emph("turso auth login"))
+		return "", fmt.Errorf("you are not logged in, please login with %s before running other commands.", internal.Emph("turso auth login"))
 	}
 
 	return settingsToken, nil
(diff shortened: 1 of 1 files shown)


haaawk (commented):
This will still prefix the message with "Error:" no?

haaawk (approved):

haaawk:
> > This will still prefix the message with "Error:" no?
> 
> So It should be like normal message instead of error?

That's my understanding.

internal/cmd/db.go
```diff
@@ -112,7 +112,7 @@ func getAccessToken(warnMultipleAccessTokenSources bool) (string, error) {
 		return envToken, nil
 	}
 	if !isJwtTokenValid(settingsToken) {
-		return "", fmt.Errorf("user not logged in, please login with %s", internal.Emph("turso auth login"))
+		return "", fmt.Printf("You are not logged in, please login with %s before running other commands.", internal.Emph("turso auth login"))
```
haaawk, line 115: Does this even compile? Why are you returning string as an error?

haaawk (changes requested):
I think returning string as an error is not the right way to fix the problem

PruthviThakor:
> I think returning string as an error is not the right way to fix the problem

yeah I get that, We souldn't change Error to Print

notrab:
I think it's fine to return a string instead of an error here. The original issue https://github.com/tursodatabase/turso-cli/issues/701 was to improve the wording by this kind of behaviour being common. Removing `error: ` prefix does that.

haaawk:
#701 is properly fixed by https://github.com/tursodatabase/turso-cli/pull/851

PR #720, opened 2023-11-19
Fix typo and sqld build link path - `turso dev`

8times4:
There was a typo in `turso dev` output along with the path being wrong for build instructions for sqld.
This PR fixes both.

Thanks,
-8x4


haaawk (approved):

haaawk:
Seems that the typo was fixed in another PR.

PR #721, opened 2023-11-20
show organization name when selecting plan

glommer:
Hopefully will make it less confusing.

### DIFF
--- a/internal/cmd/plan.go
+++ b/internal/cmd/plan.go
@@ -463,10 +463,20 @@ func promptPlanSelection(plans []turso.Plan, current string) (string, error) {
 		planNames = append(planNames, plan.Name)
 	}
 
+	settings, err := settings.ReadSettings()
+	if err != nil {
+		return "", err
+	}
+
+	var org string
+	if org = settings.Organization(); org == "" {
+		org = settings.GetUsername()
+	}
+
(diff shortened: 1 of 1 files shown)


gris (approved):

internal/cmd/plan.go
```diff
@@ -463,10 +463,20 @@ func promptPlanSelection(plans []turso.Plan, current string) (string, error) {
 		planNames = append(planNames, plan.Name)
 	}
 
+	settings, err := settings.ReadSettings()
+	if err != nil {
+		return "", err
+	}
+
+	var org string
+	if org = settings.Organization(); org == "" {
```
gris, line 472:
```suggestion
	if org := settings.Organization(); org == "" {
```
you can ditch the previous declaration and packing var declaration and assignment in the same line here.

glommer: I don't think I can, because org is reassigned within the branch. 

```
# github.com/tursodatabase/turso-cli/internal/cmd
internal/cmd/plan.go:478:80: undefined: org
```

gris: you are right. I think the following would be a bit more readable then, but ⛏️⛏️ ⛏️  
```suggestion
        org := settings.Organization()
	if org == "" {
```

PR #726, opened 2023-11-25
Updated libsql install instructions link

pxlprfct:
- Also I assume it was meant to be "To install it[...]" - so updated that too

Looks like it's a duplicate of https://github.com/tursodatabase/turso-cli/pull/725 - sorry!


avinassh:
Thanks for the PR. But this was fixed in #725

PR #742, opened 2023-12-14
Add `turso auth whoami`

avinassh:


internal/cmd/auth.go
```diff
@@ -289,7 +298,25 @@ func logout(cmd *cobra.Command, args []string) error {
...
+	}
+	fmt.Printf("%s\n", user.Username)
+	return nil
+}
+
 func verifyIfTokenIsSetInEnv(cmd *cobra.Command, args []string) error {
+	// we want to override this for `turso auth whoami`
+	if cmd.Use == whoAmICmd.Use {
+		return nil
+	}
```
avinassh, line 319: We have a top level check for `turso auth` commands which prevents usage of `auth` command if `TURSO_API_TOKEN` is set. Just for the `whoami` sub command I have added this exception.

penberg (approved):

PR #745, opened 2023-12-15
fix: update viper mod

gris:


haaawk (approved):

internal/settings/settings.go
```diff
@@ -144,6 +144,7 @@ func (s *Settings) SetAutoupdate(autoupdate string) {
 
 func (s *Settings) SetLastUpdateCheck(t int64) {
 	config := viper.GetStringMap("config")
+	fmt.Println(config)
```
haaawk, line 147: This shouldn't be here.

gris: It is just for debugging purposes :)

gris: I mean, I'm just testing here bc I can't reproduce in my local env

haaawk: I figured but we shouldn't merge it, should we?

gris: Nope!

PR #747, opened 2023-12-15
fix: read settings before persisting changes

gris:


internal/cmd/root.go
```diff
@@ -35,12 +35,13 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&noMultipleTokenSourcesWarning, "no-multiple-token-sources-warning", false, "Don't warn about multiple access token sources")
 
 	rootCmd.PersistentPostRun = func(cmd *cobra.Command, args []string) {
+		settingsV, _ := settings.ReadSettings()
```
haaawk, line 38: Please don't ignore errors

internal/cmd/root.go
```diff
@@ -35,12 +35,17 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&noMultipleTokenSourcesWarning, "no-multiple-token-sources-warning", false, "Don't warn about multiple access token sources")
 
 	rootCmd.PersistentPostRun = func(cmd *cobra.Command, args []string) {
+		configSettings, err := settings.ReadSettings()
 		settings.PersistChanges()
```
haaawk, line 39: This should be below error check

haaawk (approved):

PR #751, opened 2023-12-16
fix(cli,shell): url string using https scheme. replaced with libsql

KunalSin9h:


internal/cmd/db_show.go
```diff
@@ -14,6 +14,7 @@ import (
 func init() {
 	dbCmd.AddCommand(showCmd)
 	showCmd.Flags().BoolVar(&showUrlFlag, "url", false, "Show URL for the database HTTP API.")
+	showCmd.Flags().BoolVar(&showHttpUrlFlag, "http-url", false, "Show HTTP URL for the database HTTP API.")
```
avinassh, line 17: I am not sure why is this needed?

avinassh:
hey, I am not following the changes here. I do see some commits changing the protocol, but it is reverted now

internal/cmd/db_show.go
KunalSin9h: showHttpUrlFlag is defined bot not be register as flag in the cobra. 
https://github.com/tursodatabase/turso-cli/blob/5eeb66c637641426c94a6ee8659bc22939e9c508/internal/cmd/db.go#L17

KunalSin9h:
@avinassh i made some change to use **libsql://....** when using `db shell` command. but later found better solution for the same, so i made appropriate changes.

As discussed here https://discord.com/channels/933071162680958986/1185565974699712512

KunalSin9h:
@avinassh i am waiting for this.

penberg (approved):

avinassh (approved):

avinassh:
@KunalSin9h merged it, thank you for the contribution 🎉

PR #754, opened 2023-12-20
Improve error message for `turso db shell http://badurl:8000`

JustLeif:
PR for issue https://github.com/tursodatabase/turso-cli/issues/694

The goal being to remove the obscured `Error: failed to connect to database. err: failed to execute SQL: SELECT 1;` error, but keep the improved specific error with an error code added in https://github.com/tursodatabase/libsql-client-go/pull/81

### Examples
Screenshot from 2023-12-19 19-37-02

### DIFF
--- a/internal/cmd/db_shell.go
+++ b/internal/cmd/db_shell.go
@@ -180,9 +180,9 @@ var shellCmd = &cobra.Command{
 			shellErr := strings.Split(err.Error(), "\n")
 			var serverErr string
 			if 1 < len(shellErr) {
-				serverErr = "\n" + shellErr[1]
+				serverErr = shellErr[1]
 			}
-			return fmt.Errorf("connection to %s failed. Is this a valid URI?%s", internal.Emph(shellConfig.DbUri), serverErr)
+			return fmt.Errorf("connection to %s failed. %s", internal.Emph(shellConfig.DbUri), serverErr)
 		}
 
 		return nil
--- a/internal/cmd/db_shell.go
+++ b/internal/cmd/db_shell.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/libsql/libsql-shell-go/pkg/shell"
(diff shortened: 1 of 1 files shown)


penberg:
@athoscouto @avinassh please review

internal/cmd/db_shell.go
```diff
@@ -174,7 +175,17 @@ var shellCmd = &cobra.Command{
...
 		}
 
-		return shell.RunShell(shellConfig)
+		err := shell.RunShell(shellConfig)
+		if err != nil {
+			shellErr := strings.Split(err.Error(), "\n")
+			var serverErr string
+			if 1 < len(shellErr) {
+				serverErr = shellErr[1]
+			}
```
athoscouto, line 184: What is the reasoning behind this logic?
Note that if the error doesn't have more than one line, `serverErr` will be empty.
If you're trying to remove blank lines, you can use something like `strings.TrimSpace`.

athoscouto (changes requested):
Hi @JustLeif, thanks for the contribution!

This won't only change error messages when connecting with wrong URLs.
It will add the `connection to ...` message to any error returned from the shell invocation.

From `turso-cli` code there is no certain way to tell the difference between connection and other errors.
The best place to make this change is on that package that implements the shell, [here](https://github.com/tursodatabase/libsql-shell-go/blob/main/pkg/shell/shell.go#L36-L38).

Feel free to take a shot there.
If you want me to do it, just let me know - I'll have time to tackle this next week.

internal/cmd/db_shell.go
JustLeif: Thanks for the review! The reasoning here was to only print the string that was added in this PR: https://github.com/tursodatabase/libsql-client-go/pull/81/files
(A string that displays the specific server connection error)

I would love to work on it, but I would like to know: in a perfect world, what would the error message the user received look like when being unable to access the shell for any reason?

I didn't see any documentation about error message standards (besides the use of internal.Emph()), so let me know if there is something I should know.

Thanks again :)

PR #756, opened 2023-12-20
Add turso group update cmd

athoscouto:


internal/flags/extensions.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"github.com/spf13/cobra"
+)
+
+var extensionsFlag string
+
+func AddExtensions(cmd *cobra.Command, desc string) {
+	cmd.Flags().StringVar(&extensionsFlag, "extensions", "", desc)
+	_ = cmd.RegisterFlagCompletionFunc("extensions", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"all", "none"}, cobra.ShellCompDirectiveNoFileComp
```
gris, line 12: Should we support activating individual extensions too?

gris (approved):

internal/flags/extensions.go
athoscouto: Not here. If we decide to expose this feature we have to make it work with group/database create as well.

Issue #608, opened 2023-09-03
Useless `auth login` error message


penberg:
I don't know how to remedy the situation.

```
penberg@vonneumann api % turso auth login
Error: auth commands aren't effective when a token is set in the environment variable
```

It should at minimum tell me what environment variable is set...

notrab:
Looks like this got resolved today in https://github.com/tursodatabase/turso-cli/pull/778

PR #761, opened 2023-12-27
chore: Log which environment variable is set

p0isonra1n:
This address issue #608 to show which environment variable the token is being pulled from.


notrab:
Thanks for the PR @p0isonra1n! 🥳 

It looks like this got resolved internally today without this PR being discovered, sorry about that!

Ping me on Discord (`notrab`) so I can send you some swag as a thank you 🧦

PR #771, opened 2024-01-08
docs: update arg placeholders to use the same style

notrab:
We should adhere to a common format when describing required and optional arguments for commands.

Typically, the standards are follows:

* `<required arg>`
* `[optional option]`

But we've been using the following:

* `[database_name]`
* `database_name`
* `[group]`

I also added the required `<>` formatting to to `api-token-name` since that didn't use it.

I'm open to changing `<database-name>`/`<api-token-name>` to `<database_name>`/`<api_token_name>`. It's not clear which format we prefer.

### DIFF
--- a/internal/cmd/db_create.go
+++ b/internal/cmd/db_create.go
@@ -29,7 +29,7 @@ func init() {
 }
 
 var createCmd = &cobra.Command{
-	Use:               "create [flags] [database_name]",
+	Use:               "create [flags] [database-name]",
 	Short:             "Create a database.",
 	Args:              cobra.MaximumNArgs(1),
 	ValidArgsFunction: noFilesArg,
--- a/internal/cmd/db_replicate.go
+++ b/internal/cmd/db_replicate.go
@@ -22,7 +22,7 @@ func init() {
 }
 
 var replicateCmd = &cobra.Command{
-	Use:               "replicate <database-name> [location-id]",
+	Use:               "replicate <database-name> <location-code>,
 	Short:             "Replicate a database.",
 	Args:              cobra.RangeArgs(1, 2),
 	ValidArgsFunction: replicateArgs,
--- a/internal/cmd/db_replicate.go
+++ b/internal/cmd/db_replicate.go
@@ -22,7 +22,7 @@ func init() {
 }
 
 var replicateCmd = &cobra.Command{
-	Use:               "replicate database_name location_id",
+	Use:               "replicate <database-name> location_id",
 	Short:             "Replicate a database.",
 	Args:              cobra.RangeArgs(1, 2),
 	ValidArgsFunction: replicateArgs,
--- a/internal/cmd/db_replicate.go
+++ b/internal/cmd/db_replicate.go
@@ -22,7 +22,7 @@ func init() {
 }
 
 var replicateCmd = &cobra.Command{
-	Use:               "replicate <database-name> <location-code>,
+	Use:               "replicate <database-name> <location-code>",
 	Short:             "Replicate a database.",
 	Args:              cobra.RangeArgs(1, 2),
 	ValidArgsFunction: replicateArgs,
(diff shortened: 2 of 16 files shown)


internal/cmd/db_create.go
```diff
@@ -29,7 +29,7 @@ func init() {
 }
 
 var createCmd = &cobra.Command{
-	Use:               "create [flags] [database_name]",
+	Use:               "create [flags] [database-name]",
```
athoscouto, line 32: Should we remove all occurrences of `[flags]`?

internal/cmd/db_replicate.go
```diff
@@ -22,7 +22,7 @@ func init() {
 }
 
 var replicateCmd = &cobra.Command{
-	Use:               "replicate database_name location_id",
+	Use:               "replicate <database-name> location_id",
```
athoscouto, line 25:
```suggestion
	Use:               "replicate <database-name> [location-id]",
```

athoscouto (approved):
LGTM

internal/cmd/db_create.go
notrab: I'm 50/50 on that. I like that we include it so it hints that there are optional flags, perhaps we move it after `[database-name]`.

I'm easy with whatever is decided.

athoscouto: Almost all commands have flags, so I feel it is a bit redundant to add it in all the places.
It is also something that is easy to forget to add when there is a flag or forget to hide when there isn't any.

internal/cmd/db_replicate.go
```diff
@@ -22,7 +22,7 @@ func init() {
 }
 
 var replicateCmd = &cobra.Command{
-	Use:               "replicate database_name location_id",
+	Use:               "replicate <database-name> [location-id]",
```
notrab, line 25:
```suggestion
	Use:               "replicate <database-name> <location-code>,
```

internal/cmd/group_locations.go
```diff
@@ -51,7 +51,7 @@ var groupLocationsListCmd = &cobra.Command{
 }
 
 var groupLocationAddCmd = &cobra.Command{
-	Use:               "add [group] [...locations]",
+	Use:               "add <group-name> [...locations]",
```
notrab, line 54:
```suggestion
	Use:               "add <group-name> <location-code>",
```

internal/cmd/group_locations.go
```diff
@@ -118,7 +118,7 @@ var groupLocationAddCmd = &cobra.Command{
 }
 
 var groupsLocationsRmCmd = &cobra.Command{
-	Use:               "remove [group] [...locations]",
+	Use:               "remove <group-name> [...locations]",
```
notrab, line 121:
```suggestion
	Use:               "remove <group-name> <location-code>",
```

internal/cmd/db_replicate.go
```diff
@@ -22,7 +22,7 @@ func init() {
 }
 
 var replicateCmd = &cobra.Command{
-	Use:               "replicate database_name location_id",
+	Use:               "replicate <database-name> <location-code>,
```
notrab, line 25:
```suggestion
	Use:               "replicate <database-name> <location-code>",
```

Fixes

notrab:
@athoscouto aside from the `[flags]`, is there anything else you'd like to see changed here? I'm nervous to remove that, that's all.

Also, if I want to merge this. Do I just merge and the CLI is published automatically?

athoscouto:
@notrab feel free to merge it as it is. I suggest using "squash and merge".
To release a new version you need to tag the commit with the semver you want to release.
I think it should be `v0.88.0` is this case, since I've added the transfer feature and that hasn't been released yet.

notrab:
@athoscouto cool, thank you. I'm happy to merge as is and wait til the next tag goes out with the transfer feature (which contains the new format too `<...>` I think).

PR #775, opened 2024-01-12
turso group tokens cmd

athoscouto:
Creates a separate set of commands to handle group tokens.

I wanted to avoid doing this, but `turso db token invalidate` is misleading.
The proper solution is to have a place to manage group tokens.
And, if somebody tries to `turso db token invalidate` on a group database, we should error out.


gris (approved):
https://github.com/tursodatabase/turso-cli/pull/775/files#file-internal-cmd-expiration_flag-go-L12 maybe it would be nice to remove the dead code in this same pr

PR #789, opened 2024-02-05
feat(cmd/quickstart): open flag

### LINKED ISSUE #787: Quickstart page opened as part of auto-update?
We have a user point out that the Turso quickstart document page opens in the browser after few days of use. One thing I can think of is that our auto-update feature runs just `brew` and, therefore, runs the `turso quickstart` command. Let's verify to make sure autoupdate does not do this.

notrab:
Fixes #787

`turso quickstart` will also prompt users to open the URL to get started instead of opening automatically. Now users are prompted to pass `-o` or `--open` when running `turso quickstart`.

### DIFF
--- a/internal/cmd/quickstart.go
+++ b/internal/cmd/quickstart.go
@@ -7,8 +7,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var openBrowser bool
+
 func init() {
 	rootCmd.AddCommand(quickstartCmd)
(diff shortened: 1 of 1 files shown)


athoscouto:
I don't think this is the right way to go about this.
brew shouldn't run quickstart during updates.

notrab:
@athoscouto it made sense with the previous behavior to run `quickstart` which was to display the inline quickstart.

I guess the alternative would be to remove the post_install script. Feel free to modify the branch.

PR #790, opened 2024-02-05
Invalidate all user sessions during logout if the -a flag is used

athoscouto:
```
> turso auth logout --help
Log out currently logged in user.

Usage:
  turso auth logout [flags]

Flags:
  -a, --all    Invalidate all sessions for the current user.
  -h, --help   help for logout

Global Flags:
  -c, --config-path string   Path to the directory with config file
  ```
  

--- 
Example:

```
> turso auth logout -a
Invalidated all sessions started before 2024-02-05 15:08:11 UTC.
Logged out.
```

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -10,10 +10,12 @@ import (
 	"os"
 	"strconv"
 	"text/template"
+	"time"
 
 	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
 	"github.com/tursodatabase/turso-cli/internal"
+	"github.com/tursodatabase/turso-cli/internal/flags"
 	"github.com/tursodatabase/turso-cli/internal/settings"
 )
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: internal/cmd/auth.go [hunk 1]
```diff
@@ -346,15 +349,30 @@ func logout(cmd *cobra.Command, args []string) error {
...
+	if token := settings.GetToken(); len(token) == 0 {
 		fmt.Println("No user logged in.")
-	} else {
-		settings.SetToken("")
-		settings.SetUsername("")
-		fmt.Println("Logged out.")
+		return nil
 	}
 
+	if flags.All() {
```
gris, line 357: nit: extract all the related logic to a function

gris (approved):

internal/cmd/auth.go
athoscouto: thanks! it is way better like this

PR #799, opened 2024-02-22
Use the granular permissions in token APIs

avinassh:
e.g usage:

```sh
turso group tokens create momos --attach awake-speedball,blessed-silver-sable,super-valkyrie
```

or 

```sh
turso group tokens create momos --attach awake-speedball --attach blessed-silver-sable --attach super-valkyrie
```

For db tokens:

```sh
turso db tokens create awake-speedball --attach blessed-silver-sable --attach super-valkyrie
```

### DIFF
--- a/internal/cmd/group_tokens.go
+++ b/internal/cmd/group_tokens.go
@@ -131,7 +131,7 @@ var groupCreateTokenCmd = &cobra.Command{
 	},
 }
 
-func validateDBNames(dbNames []string) ([]string, error) {
+func validateDBNames(dbNames []string) error {
 	databases := map[string]string{}
 	for _, db := range getDatabasesCache() {
 		databases[db.Name] = db.ID
--- a/internal/cmd/group_tokens.go
+++ b/internal/cmd/group_tokens.go
@@ -86,7 +86,7 @@ func init() {
 	groupTokensCmd.AddCommand(groupCreateTokenCmd)
 	flags.AddExpiration(groupCreateTokenCmd)
 	flags.AddReadOnly(groupCreateTokenCmd)
-	flags.AddAttachClaim(groupCreateTokenCmd)
+	flags.AddAttachClaims(groupCreateTokenCmd)
 }
 
 var groupCreateTokenCmd = &cobra.Command{
@@ -112,13 +112,13 @@ var groupCreateTokenCmd = &cobra.Command{
 			return err
 		}
 		var claim *turso.PermissionsClaim
-		if len(flags.AttachClaim()) > 0 {
-			dbNames, err := validateDBNames(flags.AttachClaim())
+		if len(flags.AttachClaims()) > 0 {
+			err := validateDBNames(flags.AttachClaims())
 			if err != nil {
 				return err
 			}
 			claim = &turso.PermissionsClaim{
-				ReadAttach: turso.Entities{DBNames: dbNames},
+				ReadAttach: turso.Entities{DBNames: flags.AttachClaims()},
 			}
 		}
 		token, err := client.Groups.Token(group.Name, expiration, flags.ReadOnly(), claim)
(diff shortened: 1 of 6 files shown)


internal/turso/groups.go
```diff
@@ -190,13 +191,36 @@ func (d *GroupsClient) WaitLocation(name, location string) error {
...
+func (d *GroupsClient) Token(group string, expiration string, readOnly bool, permissions *PermissionsClaim) (string, error) {
 	authorization := ""
 	if readOnly {
 		authorization = "&authorization=read-only"
 	}
 	url := d.URL(fmt.Sprintf("/%s/auth/tokens?expiration=%s%s", group, expiration, authorization))
-	r, err := d.client.Post(url, nil)
+
+	var body io.Reader
+	if permissions != nil {
```
athoscouto, line 214: ⛏️: You can remove the if, the body will be empty if `permissions` is `nil`.

internal/cmd/group_tokens.go
```diff
@@ -110,8 +111,17 @@ var groupCreateTokenCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-
-		token, err := client.Groups.Token(group.Name, expiration, flags.ReadOnly())
+		var claim *turso.PermissionsClaim
+		if len(flags.Claim()) > 0 {
```
athoscouto, line 115: We should use a more specific name. This is not a generic claim, it is a list of databases that are allowed to read attached.

internal/cmd/group_tokens.go
```diff
@@ -120,3 +130,18 @@ var groupCreateTokenCmd = &cobra.Command{
 		return nil
 	},
 }
+
+func validateDBNames(dbNames []string) ([]string, error) {
```
athoscouto, line 134: I suggest changing the API to take database names instead, since it is how we do in most (if not all) cases.

internal/cmd/group_tokens.go
athoscouto: ⛏️: the name should be a plural as well.

haaawk:
What is `turso group tokens create momos --claim awake-speedball,blessed-silver-sable,super-valkyrie` doing?
Is it a group token that has rw access to all databases in group momos and attach access to selected dbs?

haaawk:
Is there a way for a user to get a token that has rw access to a single db and attach access to selected dbs?

internal/cmd/group_tokens.go
avinassh: > We should use a more specific name. This is not a generic claim, it is a list of databases that are allowed to read attached.

I agree. What do you think about `--attach-claims`? (`attach-read` is not needed since attach is going to read only)

athoscouto: It is better. 

I'd prefer something like `--attach` or `--attach-to`. 
Since we are in the context of a token creation command adding the `-claims` suffix doesn't add much, IMO.

But as long as it mentions attach, I'm fine! Feel free to pick the name.

avinassh: Makes sense, going with `--attach`

avinassh:
@athoscouto @haaawk  I have updated the PR and added the missing `db tokens` too. Addressed the previous PR review concerns

internal/cmd/group_tokens.go
athoscouto, line 134: ⛏️: you can simplify the function to:
```suggestion
func validateDBNames(dbNames []string) error {
```

internal/flags/claim.go
```diff
@@ -0,0 +1,13 @@
+package flags
+
+import "github.com/spf13/cobra"
+
+var attachFlag []string
+
+func AddAttachClaim(cmd *cobra.Command) {
```
athoscouto, line 7: ⛏️:
```suggestion
func AddAttachClaims(cmd *cobra.Command) {
```

internal/flags/claim.go
```diff
@@ -0,0 +1,13 @@
...
+	cmd.Flags().StringSliceVar(&attachFlag, "attach", nil, "list of database names with attach claim to be added to the token")
+}
+
+func AttachClaim() []string {
```
athoscouto, line 11: ⛏️:
```suggestion
func AttachClaims() []string {
```

athoscouto (approved):

PR #802, opened 2024-02-23
Use persistent connection for shell

avinassh:
This patch uses websocket connections instead of
HTTP for the shell. In HTTP mode, we use a connection pool. So, at every invocation in shell, can use a
different connection. While this is okay for most cases, but this doesn't work for operations involving stateful connections viz. transactions, ATTACH statements and a few PRAGMA statements.

closes:
https://github.com/tursodatabase/turso-cli/issues/784 
https://github.com/tursodatabase/libsql-shell-go/issues/163

---

## Before 

```shell
→  PRAGMA foreign_keys=OFF;
→  PRAGMA foreign_keys;
FOREIGN KEYS
1
→
→  BEGIN;
→  SELECT 42;
42
42
→  COMMIT;
Error: failed to execute SQL: COMMIT;
SQLite error: cannot commit - no transaction is active
→
```

## Afterr

```shell
→  PRAGMA foreign_keys=OFF;
→  PRAGMA foreign_keys;
FOREIGN KEYS
0
→  BEGIN;
→  SELECT 42;
42
42
→  COMMIT;
```

### DIFF
--- a/internal/cmd/db_shell.go
+++ b/internal/cmd/db_shell.go
@@ -39,10 +39,10 @@ func getURL(db *turso.Database, client *turso.Client) (string, error) {
 		}
 		for _, instance := range instances {
 			if instance.Region == locationFlag {
-				return getInstanceHttpUrl(db, &instance), nil
+				return getInstanceWSUrl(db, &instance), nil
 			}
 			if instance.Name == instanceFlag {
-				return getInstanceHttpUrl(db, &instance), nil
+				return getInstanceWSUrl(db, &instance), nil
 			}
 		}
 		if locationFlag != "" {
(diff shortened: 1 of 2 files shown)


athoscouto (approved):
LGTM because it fixes the issue.

But I'd like @haaawk to take a look here as well.
We should be able to implement a shell connection over HTTP.
Is this a limitation from our SQL Go driver or Hrana architecture?

avinassh:
> We should be able to implement a shell connection over HTTP.
> Is this a limitation from our SQL Go driver or Hrana architecture?

I think it is a bit of both. In Go, we don't send other metadata in the HTTP request. The sqld server requires some stream identifier to recognise a connection. 

I am not sure how it is done in other clients.

haaawk:
> LGTM because it fixes the issue.
> 
> But I'd like @haaawk to take a look here as well. We should be able to implement a shell connection over HTTP. Is this a limitation from our SQL Go driver or Hrana architecture?

To guarantee usage of the same SQLD db connection we would have to use the same stream. (each stream is guaranteed to use a single SQLD db connection). The problem is that HRANA streams expire and there's no way to make a stream that don't expire.

haaawk (approved):

PR #814, opened 2024-03-14
Run tests on both main and canary images

avinassh:


haaawk (approved):

athoscouto:
I'm not too fond of this idea, since we don't control the lifetime of the canary image.
Using canary on CLI integration tests means that CI can break at any time because of unreleased changes to `libsql-server`.
Even though it is useful to catch those errors, I don't think adding those tests here is the way to go.

haaawk:
> I'm not too fond of this idea, since we don't control the lifetime of the canary image. Using canary on CLI integration tests means that CI can break at any time because of unreleased changes to `libsql-server`. Even though it is useful to catch those errors, I don't think adding those tests here is the way to go.

Fine. Let's have it only on tursotests and turso.

athoscouto:
@haaawk my bad, saw the internal PR and decided to move the discussion there.
It seems that I've run into concurrency issues in both cases 😅

Issue #811, opened 2024-03-08
Allow other characters than `,` as separators while importing a CSV


avinassh:
You can create a new db from CSV:

```
turso db create --from-csv test12_sanitized.csv --csv-table-name some_table
```

but this assumes`,` as a separator character. Add an optional flag `--csv-separator` which takes the separator character like:

```
turso db create --from-csv test12_sanitized.csv --csv-table-name some_table --csv-separator ";"
```

reported here: https://discord.com/channels/933071162680958986/1215308443217563668/1215308443217563668

PR #815, opened 2024-03-14
added flag csv separator for db create command

Mr-Destructive:
Added a flag for a custom separator in the `db create` command.

Addressing  #811

Example Usage:

```
turso db create --from-csv temp.csv --csv-table-name "Employee" --csv-separator ";"
```

image

Thanks.

### DIFF
--- a/internal/cmd/from_csv_flag.go
+++ b/internal/cmd/from_csv_flag.go
@@ -1,8 +1,6 @@
 package cmd
 
-import (
-	"github.com/spf13/cobra"
-)
+import "github.com/spf13/cobra"
 
 var fromCSVFlag string
 var csvTableNameFlag string
--- a/internal/flags/csvseparator.go
+++ b/internal/flags/csvseparator.go
@@ -6,7 +6,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var csvSeparator rune
 var csvSeparatorValue string
 
 func AddCSVSeparator(cmd *cobra.Command) {
@@ -18,8 +17,7 @@ func CSVSeparator() (rune, error) {
 	if err := validateCSVSeparator(csvSeparatorValue); err != nil {
 		return 0, err
 	}
-    csvSeparator = rune(csvSeparatorValue[0])
-	return csvSeparator, nil
+	return rune(csvSeparatorValue[0]), nil
 }
 
 func validateCSVSeparator(csvSeparatorValue string) error {


internal/cmd/from_csv_flag.go
```diff
@@ -4,6 +4,7 @@ import "github.com/spf13/cobra"
 
 var fromCSVFlag string
 var csvTableNameFlag string
+var csvSeparatorFlag string
```
athoscouto, line 7: Mind adding new flags on the `flags` package?
One example would be the [readonly](https://github.com/tursodatabase/turso-cli/blob/main/internal/flags/readonly.go) flag.

internal/cmd/group_flag.go
```diff
@@ -89,7 +89,10 @@ func parseDBSeedFlags(client *turso.Client) (*turso.DBSeed, error) {
 	}
 
 	if fromCSVFlag != "" {
-		return handleCSVFile(client, fromCSVFlag, csvTableNameFlag)
+		if csvSeparatorFlag != "" {
+			return handleCSVFile(client, fromCSVFlag, csvTableNameFlag, csvSeparatorFlag)
+		}
+		return handleCSVFile(client, fromCSVFlag, csvTableNameFlag, ",")
```
athoscouto, line 95:
```suggestion
		return handleCSVFile(client, fromCSVFlag, csvTableNameFlag, csvSeparatorFlag)
```

Flag already has a default value, so no need to treat the empty case.


When you need to handle it (not the case here), I suggest doing something like:
```go
		separator := ","
		if csvSeparatorFlag != "" {
			separator = csvSeparatorFlag
		}
		return handleCSVFile(client, fromCSVFlag, csvTableNameFlag, separator)
```

internal/cmd/group_flag.go
```diff
@@ -229,7 +232,7 @@ func dumpSQLiteDatabase(database string, dump *os.File) error {
 	return nil
 }
 
-func handleCSVFile(client *turso.Client, file, csvTableName string) (*turso.DBSeed, error) {
+func handleCSVFile(client *turso.Client, file, csvTableName string, csvSeparator string) (*turso.DBSeed, error) {
```
athoscouto, line 235: ⛏️: you're inside the context of the `handleCSVFile` you can call it just `separator` with no loss on semantics. 

```suggestion
func handleCSVFile(client *turso.Client, file, csvTableName string, separator string) (*turso.DBSeed, error) {
```

internal/cmd/group_flag.go
```diff
@@ -243,7 +246,10 @@ func handleCSVFile(client *turso.Client, file, csvTableName string) (*turso.DBSe
 	}
 	defer csvFile.Close()
 
-	errs, invalid, _ := csvlint.Validate(csvFile, ',', false)
+	if len(csvSeparator) == 0 {
```
athoscouto, line 249: ⛏️: we usually just check if the string is empty.
```suggestion
	if csvSeparator == "" {
```

Nice defensive programming here!

internal/cmd/group_flag.go
```diff
@@ -243,7 +246,10 @@ func handleCSVFile(client *turso.Client, file, csvTableName string) (*turso.DBSe
...
+		csvSeparator = ","
+	}
+	errs, invalid, _ := csvlint.Validate(csvFile, rune(csvSeparator[0]), false)
```
athoscouto, line 252: Hmm, if you're going to need a single rune here, you may want to validate the separator when getting the value from the flag.

You may find an example of how to do it [here](https://github.com/tursodatabase/turso-cli/blob/main/internal/flags/expiration.go).

I suggest doing all the verification at the flag level and making the function return a rune. Then `handleCSVFile` wouldn't receive a string, but a rune as well and you wouldn't need to care about empty strings or strings with more than one character.

athoscouto (commented):
Left some comments about structure and validation.
But will let @avinassh take a look since he knows this part of the code better.

internal/cmd/group_flag.go
Mr-Destructive: yes, that seems logical, will add the validation for the flag

Mr-Destructive:
Thanks for the review, @athoscouto 
I have made the required changes, it looks a lot cleaner now.

Let me know if there are any other improvements to be made.
Thanks

internal/cmd/from_csv_flag.go
```diff
@@ -1,6 +1,8 @@
 package cmd
 
-import "github.com/spf13/cobra"
+import (
```
haaawk, line 3: Please revert and try not to upload noise like this.

internal/flags/csvseparator.go
```diff
@@ -0,0 +1,33 @@
...
+func AddCSVSeparator(cmd *cobra.Command) {
+	usage := "CSV separator character. Must be a single character. Defaults to ','"
+	cmd.Flags().StringVar(&csvSeparatorValue, "csv-separator", ",", usage)
+}
+
+func CSVSeparator() (rune, error) {
+	if err := validateCSVSeparator(csvSeparatorValue); err != nil {
+		return 0, err
+	}
+    csvSeparator = rune(csvSeparatorValue[0])
```
haaawk, line 21: Is this formatted correctly?

internal/flags/csvseparator.go
```diff
@@ -0,0 +1,33 @@
+package flags
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var csvSeparator rune
+var csvSeparatorValue string
```
haaawk, line 10: Why are both needed? It seems that csvSeparator is not really used. You could just remove it and instead in line 22 return rune(csvSeparatorValue[0]), nil

haaawk (commented):
LGTM in general but there are a few things that need cleaning up.

haaawk (approved):

avinassh (approved):

avinassh:
Thank you @Mr-Destructive!

PR #819, opened 2024-03-16
add group show command

glommer:
After group update, it is more natural to recommend group show to check the version, versus finding a db inside the group, and then checking its version.

There is no payload in the group API response with the version, so for now we just iterate over the databases in the group.

### DIFF
--- a/internal/cmd/group.go
+++ b/internal/cmd/group.go
@@ -25,8 +25,33 @@ func init() {
 	addCanaryFlag(groupsCreateCmd)
 	groupCmd.AddCommand(groupsDestroyCmd)
 	addYesFlag(groupsDestroyCmd, "Confirms the destruction of the group, with all its locations and databases.")
+	groupCmd.AddCommand(groupShowCmd)
 }
 
+var groupShowCmd = &cobra.Command{
+	Use:               "show <group-name>",
+	Short:             "Show information about a group.",
+	Args:              cobra.ExactArgs(1),
+	ValidArgsFunction: groupArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		client, err := authedTursoClient()
+		if err != nil {
+			return err
+		}
+
+		name := args[0]
+		group, err := getGroup(client, name)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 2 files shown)


internal/cmd/group.go
```diff
@@ -25,8 +25,48 @@ func init() {
...
 	groupCmd.AddCommand(groupsDestroyCmd)
 	addYesFlag(groupsDestroyCmd, "Confirms the destruction of the group, with all its locations and databases.")
+	groupCmd.AddCommand(groupShowCmd)
 }
 
+var groupShowCmd = &cobra.Command{
+	Use:               "show <group-name>",
+	Short:             "Show information about a group.",
+	Args:              cobra.ExactArgs(1),
+	ValidArgsFunction: dbNameArg,
```
athoscouto, line 35: This should be a group name instead of a db name

internal/cmd/group.go
```diff
@@ -25,8 +25,48 @@ func init() {
...
+		if err != nil {
+			fmt.Printf("group %s not found\n", internal.Emph(name))
+			return nil
+		}
+
+		version := "unknown"
+		for _, db := range databases {
+			if name == db.Group {
+				if db.Version != "" {
+					version = db.Version
```
athoscouto, line 60: We should get the version from the API, let me get that fixed for ya

internal/cmd/group.go
glommer: done

internal/cmd/group.go
glommer: done

internal/cmd/group.go
```diff
@@ -25,8 +25,34 @@ func init() {
...
+		cmd.SilenceUsage = true
+		client, err := authedTursoClient()
+		if err != nil {
+			return err
+		}
+
+		name := args[0]
+		group, err := getGroup(client, name)
+		if err != nil {
+			fmt.Printf("group %s not found\n", internal.Emph(name))
```
athoscouto, line 46: you may just return the error, we should parse any error from the API and display it correctly
printing not found may not be the correct thing in all situations

athoscouto (approved):

PR #828, opened 2024-03-21
bugfix: avoid using cache in attach commands

avinassh:
If you create a db and then try to attach immediately, it fails since attach cmds look up in the cache and cache might not be populated at that time. This patch changes this behaviour and uses the API call

### DIFF
--- a/internal/cmd/group_tokens.go
+++ b/internal/cmd/group_tokens.go
@@ -132,12 +132,25 @@ var groupCreateTokenCmd = &cobra.Command{
 }
 
 func validateDBNames(client *turso.Client, dbNames []string) error {
-	databases := map[string]struct{}{}
-	for _, dbName := range getDatabaseNames(client) {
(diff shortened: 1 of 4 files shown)


gris (approved):

internal/cmd/group_tokens.go
```diff
@@ -131,13 +131,26 @@ var groupCreateTokenCmd = &cobra.Command{
 	},
 }
 
-func validateDBNames(dbNames []string) error {
-	databases := map[string]string{}
-	for _, db := range getDatabasesCache() {
-		databases[db.Name] = db.ID
+func validateDBNames(client *turso.Client, dbNames []string) error {
```
athoscouto, line 134: ⛏️: you're doing the same logic twice here, I suggest creating a function that received the database map and db names and checks if any is missing. Then you call it with cached values and call it again with fresh if the first one failed.

athoscouto (approved):

PR #842, opened 2024-04-13
feat: list invoices cmd

gris:
Screen Shot 2024-04-13 at 8 38 09 PM

### DIFF
--- a/internal/cmd/invoice.go
+++ b/internal/cmd/invoice.go
@@ -1,0 +1,12 @@
+package cmd
+
+import "github.com/spf13/cobra"
+
+func init() {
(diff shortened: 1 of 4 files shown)


internal/cmd/invoice.go
```diff
@@ -0,0 +1,12 @@
+package cmd
+
+import "github.com/spf13/cobra"
+
+func init() {
+	rootCmd.AddCommand(invoiceCmd)
```
athoscouto, line 6: I suggest avoiding cluttering the global cmd with this and adding it to `turso org`, making it a sibling of `turso org billing`.

athoscouto (approved):

PR #845, opened 2024-04-19
feat: billing info mgmt cmds

gris:
Screen Shot 2024-04-19 at 3 22 47 PM
Screen Shot 2024-04-19 at 3 23 04 PM


athoscouto (approved):
LGTM

Have you considered using interactive prompts instead of flags?

gris:
done! much better experience indeed to use prompts

PR #851, opened 2024-05-05
Improve logged out message

### LINKED ISSUE #701: Improve logged out message

haaawk:
Fixes #701

### DIFF
--- a/internal/cmd/db.go
+++ b/internal/cmd/db.go
@@ -109,6 +109,14 @@ var dbCmd = &cobra.Command{
 
 const ENV_ACCESS_TOKEN = "TURSO_API_TOKEN"
 
+type notLoggedInError struct{}
+
+func (*notLoggedInError) Error() string {
+	return fmt.Sprintf("user not logged in, please login with %s", internal.Emph("turso auth login"))
+}
+
+var NotLoggedInError = &notLoggedInError{}
+
 func getAccessToken() (string, error) {
 	token, err := envAccessToken()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/cmd/db.go
```diff
@@ -109,6 +109,14 @@ var dbCmd = &cobra.Command{
 
 const ENV_ACCESS_TOKEN = "TURSO_API_TOKEN"
 
+type notLoggedInError struct{}
+
+func (*notLoggedInError) Error() string {
+	return fmt.Sprintf("user not logged in, please login with %s", internal.Emph("turso auth login"))
+}
+
+var NotLoggedInError = &notLoggedInError{}
```
avinassh, line 118: wouldn't following just suffice:

```suggestion
var NotLoggedInError = fmt.Errorf("user not logged in, please login with %s", internal.Emph("turso auth login"))
```

avinassh (approved):

internal/cmd/db.go
haaawk: I created a new type on purpose. See how type is not exported/public. This way I make sure no one creates by accident an error that matches NotLoggedInError.

haaawk: I'm committing the PR now but let's discuss further if necessary. We can change that in follow-up if we decide to.

PR #854, opened 2024-05-06
Improve messaging in `turso db update`

### LINKED ISSUE #732: Update when already up to date should return a friendlier message
> when I run turso db update the output says "Success!.... updated successfully" even though there was no update

We should update the cli to say you're already up to date with the latest if we can.

haaawk:
Fixes #732


avinassh (approved):

avinassh:
related one if you want to pick up next: https://github.com/tursodatabase/turso-cli/issues/753

PR #865, opened 2024-05-07
feat: IsSchema and Schema info

### LINKED ISSUE #824: Show schema details in turso db show
Turso db show should include the details about schema / multi dbs. 

Something like:

```
$ turso db show masterdb

...
Type: Schema
```

```
$ turso db show childdb

...
Schema: masterdb
```

notrab:
Closes #824 

I would prefer to use `Type` to return `Schema`, but our API uses `type` for `logical` so it's a bit confusing to use the same fields for different meanings. We should update the API if we want to make this change here.

We could though do this now if people prefer?

```go
if db.IsSchema {
        fmt.Println("Type: Schema")
    } else {
        fmt.Println("Type: Logical")
    }
```

### DIFF
--- a/internal/cmd/db_show.go
+++ b/internal/cmd/db_show.go
@@ -94,6 +94,10 @@ var showCmd = &cobra.Command{
 		fmt.Println("Size:          ", humanize.Bytes(dbUsage.Usage.StorageBytesUsed))
 		fmt.Println("Sleeping:      ", formatBool(db.Sleeping))
 		fmt.Println("Bytes Synced:  ", humanize.Bytes(dbUsage.Usage.BytesSynced))
+		fmt.Println("Is Schema:     ", formatBool(db.IsSchema))
+		if db.Schema != "" {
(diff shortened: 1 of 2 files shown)


internal/cmd/db_show.go
```diff
@@ -94,6 +94,10 @@ var showCmd = &cobra.Command{
 		fmt.Println("Size:          ", humanize.Bytes(dbUsage.Usage.StorageBytesUsed))
 		fmt.Println("Sleeping:      ", formatBool(db.Sleeping))
 		fmt.Println("Bytes Synced:  ", humanize.Bytes(dbUsage.Usage.BytesSynced))
+		fmt.Println("Is Schema:     ", formatBool(db.IsSchema))
+		if db.Schema != "" {
+			fmt.Println("Parent DB ID:  ", db.Schema)
```
notrab, line 99: Or

```suggestion
			fmt.Println("Schema:  ", db.Schema)
```

PR #868, opened 2024-06-05
feat(auth): add auth support in turso dev

guergabo:
This PR adds auth support for turso dev command and adds two additional checks when starting and killing the sqld server process. 

## Changes 
The `Start` method launches the `sqld` command without waiting for the process to fully initialize. If `sqld` exits with a non-zero status code due to validation issues or other errors, the error message may be missed by the user since the program prints "sqld is listening" and then blocks the channel.

Calling the `Wait` method immediately after `Start` would cause the program to hang in the happy path. Instead, this PR waits for the server to be ready by pinging the opened connection before printing the success message. It also waits after calling `Kill` to ensure a clean shutdown, as `Kill` does not wait for the process to fully exit.

Lastly, the original purpose of this contribution was to fix https://github.com/tursodatabase/turso-cli/issues/752

#### With value JWT Key
Screen Shot 2024-06-05 at 11 47 31 AM

#### With invalid JWT key
Screen Shot 2024-06-05 at 11 53 37 AM

### DIFF
--- a/internal/cmd/dev.go
+++ b/internal/cmd/dev.go
@@ -2,12 +2,14 @@ package cmd
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"os/exec"
 	"os/signal"
 	"path/filepath"
 	"regexp"
 	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/tursodatabase/turso-cli/internal"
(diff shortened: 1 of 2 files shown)


avinassh:
Hey! Thank you for the PR. If the user is passing JWT file, we should not print: 

> No auth token is required when sqld is running locally

guergabo:
Hi @avinassh, thanks for feedback. I updated the PR to match the the behavior `devFile` flag. Let me know what you think.

avinassh (approved):

avinassh:
Thank you for your contribution @guergabo! 🥳

PR #872, opened 2024-06-11
allow dev when logged out

notrab:
`turso dev` should work when signed out. This PR enables that behaviour.

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -30,7 +30,7 @@ func Execute() {
 var noMultipleTokenSourcesWarning bool
 
 func requiresLogin(cmd *cobra.Command) bool {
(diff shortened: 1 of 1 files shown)


avinassh (approved):

internal/cmd/root.go
```diff
@@ -30,7 +30,20 @@ func Execute() {
...
+		"turso update":           {},
+		"turso dev":              {},
+	}
+
+	_, excluded := excludedCommands[cmd.Name()]
+	if !excluded {
+		_, excluded = excludedCommands[cmd.CommandPath()]
+	}
+
+	return !excluded
```
avinassh, line 46: something doesn't look right here

athoscouto:
Already implemented on `main` by https://github.com/tursodatabase/turso-cli/commit/78b34c7a04d3733ca0e42aba84d9a98acaa63461.

PR #875, opened 2024-06-26
feat: list upcoming invoices

gris:


internal/cmd/invoice_flag.go
```diff
@@ -0,0 +1,9 @@
+package cmd
+
+import "github.com/spf13/cobra"
+
+var invoiceType string
+
+func AddInvoiceType(cmd *cobra.Command) {
```
athoscouto, line 7: Please add new flags on the `flags` package. You can use the [timeline](https://github.com/tursodatabase/turso-cli/blob/main/internal/flags/timeline.go#L0-L1) flag as reference. It has autocomplete and validation built in.

internal/cmd/invoice_list.go
```diff
@@ -23,7 +24,15 @@ var listInvoicesCmd = &cobra.Command{
 			return err
 		}
 
-		invoices, err := client.Invoices.List()
+		if invoiceType == "" {
+			invoiceType = "issued"
```
athoscouto, line 28: You can set `"issued"` as the default value for the flag on its configuration instead of `""`

internal/cmd/invoice_list.go
```diff
@@ -23,7 +24,15 @@ var listInvoicesCmd = &cobra.Command{
...
+		}
+
+		if invoiceType != "all" && invoiceType != "upcoming" && invoiceType != "issued" {
```
athoscouto, line 31: Better to have this validation in the flag itself

internal/cmd/invoice_list.go
```diff
@@ -46,25 +55,29 @@ func printInvoiceListTable(invoices []turso.Invoice) {
...
 	data := make([][]string, len(invoices))
 	for i, invoice := range invoices {
-		data[i] = []string{invoice.Number, invoice.Amount, invoice.DueDate, invoice.PaidAt, invoice.PaymentFailedAt}
+		data[i] = []string{invoice.Number, invoice.Amount, invoice.Status, invoice.DueDate, invoice.PaidAt, invoice.PaymentFailedAt}
 	}
 	return headers, data
 }
 
-func invoiceListPdfs(invoices []turso.Invoice) ([]string, [][]string) {
+func invoiceListLinks(invoices []turso.Invoice) ([]string, [][]string) {
```
athoscouto, line 72: ⛏️⛏️⛏️: `s/invoiceList/invoice/g` on function names. Same thing for `invoiceType` below. Feel free to ignore.

athoscouto (approved):

internal/cmd/invoice_flag.go
gris: thanks for this pointer, very useful!

PR #883, opened 2024-06-26
refactor: invoice list

gris:


internal/flags/invoice_type.go
```diff
@@ -0,0 +1,35 @@
...
+func InvoiceType() (string, error) {
+	if err := validateInvoiceType(invoiceType); err != nil {
+		return "", err
+	}
+	return invoiceType, nil
+}
+
+func validateInvoiceType(invoiceType string) error {
+	switch invoiceType {
+	case "issued", "all", "upcoming", "":
```
athoscouto, line 26: Why are you accepting `""`? Do we need to replace it with `issued`"?

internal/flags/invoice_type.go
```diff
@@ -0,0 +1,35 @@
...
+		return nil
+	default:
+		return fmt.Errorf("type parameter must be either 'all' or 'upcoming' or 'issued'")
```
athoscouto, line 29: ⛏️ 
```suggestion
		return fmt.Errorf("type parameter must be one of 'all', 'upcoming', or 'issued'")
```

athoscouto (approved):

internal/flags/invoice_type.go
gris: just being super safe. happy to remove it

PR #890, opened 2024-07-16
Signal to the shell when using schema db

haaawk:
This will allow the shell to use libsql-client-go in nontransactional batch mode because schema dbs don't support transactions


haaawk:
Depends on https://github.com/tursodatabase/libsql-shell-go/pull/176

avinassh (approved):

sivukhin (approved):
We need upgrade `libsql-shell-go` package, I guess

PR #897, opened 2024-07-30
feat: db list group and schema filters

notrab:
This adds the following flags for `db list`:

* `--group <group-name>`
```bash
go run cmd/turso/main.go db list --group vectors
NAME                 URL
mydb                 libsql://mydb-notrab.turso.io
super-fantomette     libsql://super-fantomette-notrab.turso.io
tutorial             libsql://tutorial-notrab.turso.io

```

* `--schema <schema-database-name>`
```bash
go run cmd/turso/main.go db list --schema parent-db
NAME             URL
child-db         libsql://child-db-notrab.turso.io
child-db-2       libsql://child-db-2-notrab.turso.io
new-child-db     libsql://new-child-db-notrab.turso.io
```

### DIFF
--- a/internal/cmd/db_list.go
+++ b/internal/cmd/db_list.go
@@ -8,8 +8,13 @@ import (
 	"golang.org/x/exp/slices"
 )
 
+var groupFilter string
+var schemaFilter string
+
 func init() {
 	dbCmd.AddCommand(listCmd)
+	listCmd.Flags().StringVarP(&groupFilter, "group", "g", "", "Filter databases by group")
+	listCmd.Flags().StringVarP(&schemaFilter, "schema", "s", "", "Filter databases by schema")
 }
 
 var listCmd = &cobra.Command{
@@ -24,7 +29,15 @@ var listCmd = &cobra.Command{
 			return err
 		}
 
-		databases, err := client.Databases.List()
+		options := make(map[string]string)
+		if groupFilter != "" {
+			options["group"] = groupFilter
+		}
+		if schemaFilter != "" {
+			options["schema"] = schemaFilter
+		}
+
+		databases, err := client.Databases.List(options)
 		if err != nil {
 			return err
 		}
--- a/internal/turso/databases.go
+++ b/internal/turso/databases.go
@@ -27,19 +27,18 @@ type Database struct {
 
 type DatabasesClient client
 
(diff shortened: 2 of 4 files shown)


internal/turso/databases.go
```diff
@@ -26,8 +27,22 @@ type Database struct {
 
 type DatabasesClient client
 
-func (d *DatabasesClient) List() ([]Database, error) {
-	r, err := d.client.Get(d.URL(""), nil)
+func (d *DatabasesClient) List(options ...map[string]string) ([]Database, error) {
```
haaawk, line 30: Why not just `options map[string]string`?

haaawk (approved):

internal/turso/databases.go
notrab: It required updating elsewhere we use `List` to pass `nil`. I was just trying to find a good way to make this optional.

haaawk: I see.

haaawk: Maybe add a comment to explain the trickery

notrab: Update: Reverted to use `nil`

internal/turso/databases.go
```diff
@@ -26,8 +27,21 @@ type Database struct {
 
 type DatabasesClient client
 
-func (d *DatabasesClient) List() ([]Database, error) {
-	r, err := d.client.Get(d.URL(""), nil)
+func (d *DatabasesClient) List(options map[string]string) ([]Database, error) {
```
athoscouto, line 30: When possible, default to typed options. Using a struct here that has `Group` and `Schema` fields could be a good option. You could also add an encode method to it to encapsulate the logic.

internal/cmd/db_list.go
```diff
@@ -8,8 +8,13 @@ import (
 	"golang.org/x/exp/slices"
 )
 
+var groupFilter string
+var schemaFilter string
+
 func init() {
 	dbCmd.AddCommand(listCmd)
+	listCmd.Flags().StringVarP(&groupFilter, "group", "g", "", "Filter databases by group")
```
athoscouto, line 16: For all new flags you create, use the flags package. I'll make a PR adding the group and schema flags there and add you as a reviewer.

internal/turso/databases.go
notrab: Thanks! Will PR the changes

PR #898, opened 2024-07-30
feat(auth): do not wait indefinitely for callback

### LINKED ISSUE #741: Do not wait indefinitely for login callback
Here: https://github.com/tursodatabase/turso-cli/blob/5f6edd8/internal/cmd/auth.go#L169-L183

We wait till eternity for the token. This can be problematic in headless environments. I suggest adding a timeout (say 10 mins) and failing.

notrab:
Fixes #741

We might want to tweak the error message.

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -339,7 +339,7 @@ func (a authCallback) Result() (string, error) {
 		return result, nil
 	case <-time.After(10 * time.Minute):
 		_ = a.server.Shutdown(context.Background())
-		return "", fmt.Errorf("authentication timed out, try again.")
+		return "", fmt.Errorf("authentication timed out, try again")
 	}
 }
 
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -201,7 +201,10 @@ func auth(cmd *cobra.Command, path string) error {
 	fmt.Println(url)
 	fmt.Println("Waiting for authentication...")
 
-	jwt := callbackServer.Result()
+	jwt, err := callbackServer.Result()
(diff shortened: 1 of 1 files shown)


internal/cmd/auth.go
```diff
@@ -201,7 +201,10 @@ func auth(cmd *cobra.Command, path string) error {
 	fmt.Println(url)
 	fmt.Println("Waiting for authentication...")
 
-	jwt := callbackServer.Result()
+	jwt, err := callbackServer.Result()
+	if err != nil {
+		return suggestHeadless(cmd, err)
```
notrab, line 206: We could instead just `return nil` here and not show the headless warning.

haaawk (approved):

internal/cmd/auth.go
```diff
@@ -329,10 +332,15 @@ func authCallbackServer(state string) (authCallback, error) {
...
-func (a authCallback) Result() string {
-	result := <-a.ch
-	_ = a.server.Shutdown(context.Background())
-	return result
+func (a authCallback) Result() (string, error) {
+	select {
+	case result := <-a.ch:
+		_ = a.server.Shutdown(context.Background())
+		return result, nil
+	case <-time.After(10 * time.Minute):
```
avinassh, line 340: we can reduce to 5 mins too I guess.

avinassh (approved):

internal/cmd/auth.go
notrab, line 340:
```suggestion
	case <-time.After(5 * time.Minute):
```

Sure sounds good to me

PR #900, opened 2024-07-30
feat(db/create): ask to wake sleeping groups before create

### LINKED ISSUE #894: Improve error message for sleeping databases
If a database (or actually the group) is sleeping, you get the following error message:

```console
$ turso db create foo
Error: could not create database foo: cannot create database on group default because it is sleeping
```

Let's add information on _how_ to wake it up. For example, with an URL to documentation.

notrab:
Fixes #894

This does however use a mix of the `sleeping` and `archived` terminology. We should clean that up when we know which word we want to go with.

### DIFF
--- a/internal/cmd/db_create.go
+++ b/internal/cmd/db_create.go
@@ -55,6 +55,27 @@ var createCmd = &cobra.Command{
 			return err
 		}
 
+		groupInfo, err := getGroup(client, group)
+		if err != nil {
+			return fmt.Errorf("failed to get group info: %w", err)
+		}
+
+		groupStatus := aggregateGroupStatus(groupInfo)
+		if groupStatus == "Archived 💤" {
+			fmt.Printf("The group %s is currently archived. Do you want to wake it up? [Y/n]: ", internal.Emph(group))
+			var response string
+			fmt.Scanln(&response)
+			if response == "" || response == "y" || response == "yes" {
+				err = unarchiveGroup(client, group)
+				if err != nil {
+					return fmt.Errorf("failed to wake up group: %w", err)
+				}
+				fmt.Printf("Group %s has been woken up.\n", internal.Emph(group))
+			} else {
+				return fmt.Errorf("cannot create a database in an archived group. Please wake up the group first")
+			}
+		}
+
 		location, err := locationFromFlag(client)
 		if err != nil {
 			return err


notrab:
The original issue #894 also stated that `db shell` would return the same error. We can use the method there, but I'm not sure where is best considering `db shell` can be used with local URL, and the groups concept doesn't yet apply there...

```go
spinner.Stop()
			awake, err := ensureGroupAwake(client, db.Group)
			if err != nil {
				return err
			}
			if !awake {
				return fmt.Errorf("cannot connect to a database in an archived group. Please wake up the group first")
			}
			spinner.Start()
```

notrab:
I pushed a commit to `db_shell.go` where I believe is the right place to check, but we can remove that if it's not correct.

internal/cmd/group_utils.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func ensureGroupAwake(client *turso.Client, groupName string) (bool, error) {
+	group, err := getGroup(client, groupName)
+	if err != nil {
+		return false, fmt.Errorf("failed to get group info: %w", err)
+	}
+
+	groupStatus := aggregateGroupStatus(group)
+	if groupStatus != "Archived 💤" {
```
notrab, line 18: I'm not a huge fan of checking a string with an emoji here, could lead to lots of issues. Do we need the emoji now that we've changed Sleeping to Archived where this was originally shown?

internal/cmd/group_utils.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/tursodatabase/turso-cli/internal"
+	"github.com/tursodatabase/turso-cli/internal/turso"
+)
+
+func ensureGroupAwake(client *turso.Client, groupName string) (bool, error) {
```
athoscouto, line 11: I'm not a fan of this approach because:
- We add an extra call on all invocations to handle an edge case that happens once in a while
- We have to keep track of all ops that are affected by archiving and remember to add this logic to them (e.g. rotating creds, and deleting dbs also doesn't work when the group is archived)

The "better" way would be to be able to detect archived errors from API responses and handle those properly. 

Not a blocker though, feel free to move forward if you don't have the time to address this.

internal/cmd/group_utils.go
notrab: Removed it

internal/cmd/group_utils.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		return false, fmt.Errorf("failed to get group info: %w", err)
+	}
+
+	groupStatus := aggregateGroupStatus(group)
+	if groupStatus != "Archived" {
+		return true, nil
+	}
+
+	fmt.Printf("The group %s is currently archived. Do you want to wake it up? [Y/n]: ", internal.Emph(groupName))
```
athoscouto, line 22: We already have a function to handle this logic, named `promptConfirmation`.
While you're at it if you can move it to the `prompt` package I would really appreciate it 😄

athoscouto (approved):

internal/cmd/group_utils.go
notrab: I agree this would be better.

internal/cmd/group_utils.go
```diff
@@ -0,0 +1,52 @@
...
+
+func RetryOnSleepingGroup(client *turso.Client, group string, mainSpinnerText string, action func() (any, error)) (any, error) {
+	mainSpinner := prompt.Spinner(mainSpinnerText)
+	result, actionErr := action()
+	if actionErr == nil {
+		return result, nil
+	}
+
+	errMsg := actionErr.Error()
+	if strings.Contains(errMsg, "group_sleeping") ||
```
notrab, line 21: I fear `group_sleeping` will never be reached because we don't actually return or formalise the API error from the `turso` client.

internal/cmd/group_utils.go
```diff
@@ -0,0 +1,52 @@
...
+		(strings.Contains(errMsg, "cannot create database on group") && strings.Contains(errMsg, "because it is archived")) {
```
notrab, line 22: This will be reached, but I hate having to pass in the string and detect errors like this. I would much prefer to use `group_sleeping` from the error response `code` field.

PR #901, opened 2024-07-30
refactor db list options

notrab:
This addresses feedback from @athoscouto to type and encode the query params.

### DIFF
--- a/internal/turso/databases.go
+++ b/internal/turso/databases.go
@@ -25,6 +25,8 @@ type Database struct {
 	IsSchema      bool `json:"is_schema"`
 }
 
+type DatabasesClient client
+
 type DatabaseListOptions struct {
 	Group  string
 	Schema string
@@ -40,8 +42,6 @@ func (o DatabaseListOptions) Encode() string {
 	}
 	return query.Encode()
 }
-
-type DatabasesClient client
 
 func (d *DatabasesClient) List(options DatabaseListOptions) ([]Database, error) {
 	path := d.URL("")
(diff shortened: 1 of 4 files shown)


internal/turso/databases.go
```diff
@@ -25,20 +25,30 @@ type Database struct {
...
+type DatabasesClient client
+
+func (d *DatabasesClient) List(options DatabaseListOptions) ([]Database, error) {
+	path := d.URL("")
 
-	if len(query) > 0 {
-		path += "?" + query.Encode()
+	encodedOptions := options.Encode()
+	if encodedOptions != "" {
+		path += "?" + encodedOptions
```
athoscouto, line 51: ⛏️: totally style. If you're into one-liners and reusing names 😎 
```suggestion
	if options := options.Encode(); options != "" {
		path += "?" + options
```

internal/turso/databases.go
```diff
@@ -25,20 +25,30 @@ type Database struct {
 	IsSchema      bool `json:"is_schema"`
 }
 
-type DatabasesClient client
-
-func (d *DatabasesClient) List(options map[string]string) ([]Database, error) {
-	path := d.URL("")
+type DatabaseListOptions struct {
+	Group  string `json:"group,omitempty"`
+	Schema string `json:"schema,omitempty"`
```
athoscouto, line 30: ⛏️: feel free to drop the annotations, since we're not encoding/decoding JSON.
```suggestion
	Group  string
	Schema string
```

athoscouto (approved):

PR #902, opened 2024-08-06
Replace `Sleeping` with `Archived`

avinassh:
## Description

Some users are getting confused with the terminology `Sleeping` as we also have cold starts. This patch changes the word to `Archived`. This also changes the word `Wake up` to `Unarchive`.

Previously, we had a command `turso group wakeup <group name>` which is now replaced with `turso group unarchive <group name>`

Internal slack discussion: [1](https://turso.slack.com/archives/C02J5GYN8NB/p1721980720846099), [2](https://turso.slack.com/archives/C02J5GYN8NB/p1722867218607089)

### DIFF
--- a/internal/cmd/db_wakeup.go
+++ b/internal/cmd/db_wakeup.go
@@ -15,8 +15,8 @@ func init() {
 }
 
 var wakeUpDbCmd = &cobra.Command{
-	Use:               "wakeup <db-name>",
-	Short:             "Wake up a database",
+	Use:               "unarchive <db-name>",
+	Short:             "Unarchive a database",
 	Args:              cobra.ExactArgs(1),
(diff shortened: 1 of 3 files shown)


internal/cmd/db_wakeup.go
```diff
@@ -15,8 +15,8 @@ func init() {
 }
 
 var wakeUpDbCmd = &cobra.Command{
-	Use:               "wakeup <db-name>",
```
athoscouto, line 18: Please keep the old command around as well, preferably hidden. You can use an alias to achieve that

athoscouto (approved):

internal/cmd/db_wakeup.go
avinassh: Good idea, I have updated it.

PR #907, opened 2024-08-20
redirect plan operations to dashboard

gris:
redirect 

```
turso org billing
turso plan select
turso plan upgrade
turso plan overages enable
turso plan overages disable
```

to https://app.turso.tech/$ORG/settings/billing

also, updates action versions

### DIFF
--- a/.github/workflows/brew-test.yml
+++ b/.github/workflows/brew-test.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Set up Homebrew
         id: set-up-homebrew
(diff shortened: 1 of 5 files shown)


athoscouto (commented):
I wouldn't say I like the idea of replacing all billing-related commands with opening the web app. This adds a breaking change that is avoidable.

If we don't want plan management on the CLI anymore, I suggest taking the following approach:

- Hide the current commands, keeping them working as they are today but not advertising them. The exception would be the billing portal, which has a similar behavior today.
- Create a new and more specific command to manage your turso plan, this would behave the same way as `turso org billing`, opening up the web app. I suggest `turso plan manage`. 

I don't see a reason to remove plan management logic from the CLI though - which may be totally on me since I'm not following billing-related discussions. If we still want to keep them just changing the behavior of the billing command should be enough.

athoscouto (approved):

PR #922, opened 2024-11-10
refactor: use http instead of websocket in shell by default

thilinajayanath:
- use http instead of websocket in shell by default


haaawk:
This PR does nothing.

PR #924, opened 2024-11-27
Glauber/aws improvements

glommer:
Improve AWS / Fly communication

* do not use the -l flag anymore, as it is meaningless on AWS.
* Clearly separate AWS regions from Fly regions, so the fact that both exist is not lost by accident.


avinassh:
merged via #929

PR #941, opened 2025-03-06
remove account commands

glommer:
Nobody uses those, and they add complexity


notrab:
@glommer perhaps we can keep the command but it opens tur.so/chat?

PR #948, opened 2025-03-17
Add `turso db import` command for importing SQLite files

penberg:
The functionality is the same as with `turso db create --from-file` but the command is more discoverable.

With this command, you can do:

```
$ sqlite3 hello.db
SQLite version 3.49.1 2025-02-18 13:38:58
Enter ".help" for usage hints.
sqlite> PRAGMA journal_mode = WAL;
wal
sqlite> CREATE TABLE t(x);
sqlite> INSERT INTO t(x);
sqlite> INSERT INTO t VALUEs ('Hello, world!');
sqlite>
$ turso db import hello.db
Created database hello at group aws in 1.953s.

Start an interactive SQL shell with:

   turso db shell hello

To see information about the database, including a connection URL, run:

   turso db show hello

To get an authentication token for the database, run:

   turso db tokens create hello
```

### DIFF
--- a/internal/cmd/db_import.go
+++ b/internal/cmd/db_import.go
@@ -1,0 +1,35 @@
+package cmd
+
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	dbCmd.AddCommand(importCmd)
+}
+
+var importCmd = &cobra.Command{
+	Use:               "import [filename]",
+	Short:             "Import a SQLite database file to Turso.",
+	Args:              cobra.MaximumNArgs(1),
+	ValidArgsFunction: noFilesArg,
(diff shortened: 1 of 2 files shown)


internal/cmd/db_import.go
```diff
@@ -0,0 +1,33 @@
...
+		fromFileFlag = filename
+		name := sanitizeDatabaseName(filename)
+		return CreateDatabase(name)
+	},
+}
+
+// Sanitize a SQLite database filename to be used as a Turso database name.
+func sanitizeDatabaseName(filename string) string {
+	return strings.ReplaceAll(filename, ".db", "")
+}
```
avinassh, line 33: we need to remove `/` as well if someone gives a full path? this also could fail if the name has special characters

we could also check in local cache if a db by this name exists already. if it does, we could append some random characters

avinassh (approved):

PR #949, opened 2025-03-17
Fix bug 'not being able to create db in existing fly group' and improve errors

jussisaurio:
Free tier user was not able to create a DB in an existing fly group after AWS rollout. Logic after this PR:

- If location flag is NOT provided, and user has no groups, defaults to `closestLocation`. `isValidLocation` check is made.

- If location flag is NOT provided, and user has groups, defaults to `group.Primary`. `isValidLocation` check is NOT made.

- If location flag IS provided, and user has no groups, uses that location. `isValidLocation` check is made.

- If location flag IS provided, and user HAS groups, uses that location, but the group must also have that location. `isValidLocation` check is NOT made.

Also improve the error messages a bit

### DIFF
--- a/internal/cmd/db_create.go
+++ b/internal/cmd/db_create.go
@@ -66,13 +66,18 @@ func CreateDatabase(name string) error {
 		return err
 	}
 
-	group, err := groupFromFlag(client)
+	groups, err := client.Groups.List()
+	if err != nil {
+		return err
+	}
+
+	group, err := groupFromFlag(groups)
 	if err != nil {
 		return err
 	}
 	groupName := group.Name
 
-	location, err := locationFromFlag(client)
+	location, err := locationFromFlag(client, group, groups)
 	if err != nil {
 		return err
 	}
@@ -88,7 +93,7 @@ func CreateDatabase(name string) error {
 		version = "canary"
 	}
 
-	if err := ensureGroup(client, groupName, location, version); err != nil {
+	if err := ensureGroup(client, groupName, groups, location, version); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


glommer (approved):

internal/cmd/db_create.go
```diff
@@ -175,22 +180,34 @@ func groupExists(groups []turso.Group, name string) bool {
...
+		for _, l := range group.Locations {
+			if l == loc {
+				groupContainsLocation = true
+			}
+		}
+		if !groupContainsLocation {
+			return "", fmt.Errorf("location '%s' is not valid for group '%s'. The group has the following locations: %v. You can use 'turso group locations add' to add a new location to the group", loc, group.Name, strings.Join(group.Locations, ", "))
+		}
 	}
 	if !isValidLocation(client, loc) {
```
jussisaurio, line 204: Actually this still needs a check not to fail for free tier users...........

internal/cmd/db_create.go
```diff
@@ -137,15 +142,15 @@ func getDatabaseName(args []string) (string, error) {
...
+func groupFromFlag(groups []turso.Group) (turso.Group, error) {
 
 	if groupFlag != "" {
 		if !groupExists(groups, groupFlag) {
-			return turso.Group{}, fmt.Errorf("group %s does not exist", groupFlag)
+			groupNames := make([]string, len(groups))
+			for i, g := range groups {
+				groupNames[i] = g.Name
+			}
+			return turso.Group{}, fmt.Errorf("group %s does not exist. You have the following groups: %v. You can use 'turso group create' to create a new group", groupFlag, strings.Join(groupNames, ", "))
```
sivukhin, line 153: List can be pretty large here, no?

sivukhin (approved):

internal/cmd/db_create.go
jussisaurio: Are you suggesting it can be so large that we should truncate it?

sivukhin: I think we can keep previous logic, no? Where we didn't show the list in the error message

PR #951, opened 2025-03-17
Fix bug not being able to upgrade from 0.99.x to 0.100.x

jussisaurio:


glommer (approved):

glommer:
this needs to be tagged with an upgradable-to version, though.

jussisaurio:
> Is that enough? If you are on 0.99 you will never get upgraded. This needs to be a 0.99 one I think

hahah yes...

PR #955, opened 2025-03-20
warn users if they are trying to upload a dump

glommer:
This is a common mistake, so worth checking explicitly.

### DIFF
--- a/internal/cmd/group_flag.go
+++ b/internal/cmd/group_flag.go
@@ -190,10 +190,38 @@ func countFlags(flags ...string) (count int) {
 }
 
 const MaxAWSDBSizeBytes = 1024 * 1024 * 1024 * 20 // 20 GB
+func checkIfDump(filename string) (bool, error) {
+	file, err := os.Open(filename)
+	if err != nil {
+		return false, err
+	}
+	defer file.Close()
+	scanner := bufio.NewScanner(file)
+	if scanner.Scan() {
+		firstLine := scanner.Text()
+		return strings.TrimSpace(firstLine) == "PRAGMA foreign_keys=OFF;", nil
+	} else {
+		return false, scanner.Err()
+	}
+}
+
 func sqliteFileIntegrityChecks(file string) error {
 	if flags.Debug() {
 		log.Printf("Running integrity checks on database file %s", file)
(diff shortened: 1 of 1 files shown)


internal/cmd/group_flag.go
```diff
@@ -190,10 +190,46 @@ func countFlags(flags ...string) (count int) {
...
+func checkIfDump(filename string) (bool, error) {
+	file, err := os.Open(filename)
+	if err != nil {
+		return false, err
+	}
+	defer file.Close()
+	scanner := bufio.NewScanner(file)
+	if scanner.Scan() {
+		firstLine := scanner.Text()
+		if strings.TrimSpace(firstLine) == "PRAGMA foreign_keys=OFF;" {
```
haaawk, line 202: Why not `return strings.TrimSpace(firstLine) == "PRAGMA foreign_keys=OFF;", nil`?

internal/cmd/group_flag.go
```diff
@@ -190,10 +190,46 @@ func countFlags(flags ...string) (count int) {
...
+			return true, nil
+		} else {
+			return false, nil
+		}
+	} else {
+		if err := scanner.Err(); err != nil {
```
haaawk, line 208: Why not `return false, scanner.Err()`?

haaawk (approved):

internal/cmd/group_flag.go
```diff
@@ -190,10 +190,38 @@ func countFlags(flags ...string) (count int) {
 }
 
 const MaxAWSDBSizeBytes = 1024 * 1024 * 1024 * 20 // 20 GB
+func checkIfDump(filename string) (bool, error) {
```
sivukhin, line 193: Should we instead check that file is SQLite db (by simply checking sqlite magic header string)?

sivukhin (approved):

internal/cmd/group_flag.go
glommer: We can do that, certainly. But if it *isn't* a sqlite db, then what is it? The message "this is a dump" is more helpful to users than "this is not a valid database". An incredible amount of people confuse the two!

PR #959, opened 2025-03-27
delete protection

glommer:
Allow groups and databases to be delete-protected


notrab (approved):
Copilot had comments... so lgtm! 😆

PR #968, opened 2025-04-08
feat: list and paginate databases/branches

notrab:
Blocked until https://github.com/tursodatabase/turso/pull/2644 lands.

This PR adds listing branches with a new `--branches` flag for `db show`, and some updates to including pagination.

```bash
# List databases with a limit
turso db list --limit 10

# Get the next page of databases using the cursor
turso db list --cursor eyJmIjoiSUQiLCJ2IjoxMCwicyI6MTB9

# List branches with a limit
turso db show mydb --branches --limit 5

# Get the next page of branches
turso db show mydb --branches --cursor eyJmIjoiSUQiLCJ2IjoxMCwicyI6NX0
```

### DIFF
--- a/internal/cmd/db.go
+++ b/internal/cmd/db.go
@@ -15,6 +15,7 @@ var (
 	showHttpUrlFlag      bool
 	showInstanceUrlsFlag bool
 	showInstanceUrlFlag  string
+	showBranchesFlag     bool
 	paginationLimit      int
 	paginationCursor     string
 )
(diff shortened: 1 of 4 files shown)


glommer:
I find the cursor weird. I'd just store the cursor in the cache, and add a `--next` option. Alternatively, not even store it in the cache, but use a component that allow you to press next within a session, similar to the `less` utility.

notrab:
@glommer I find adding `--next` even more weird, since you might want to show results from a specific cursor, if you know it.

I can remove the cursor pagination altogether until we settle on the DX, but right now we don't paginate the CLI results at all (they're cut off at 1000 records) but I wanted to include it in branches since it's already present.

glommer (approved):

PR #971, opened 2025-04-22
feat: add audit-logs command

notrab:


internal/turso/utils.go
```diff
@@ -24,10 +24,38 @@ func marshal(data interface{}) (io.Reader, error) {
...
+	if len(body) == 0 {
+		return nil
+	}
+
+	var errResp ErrorResponseDetails
+	if err := json.Unmarshal(body, &errResp); err != nil {
+		return fmt.Errorf("failed to unmarshal error response: %w", err)
+	}
+
+	if errResp.Code == "feature_not_available_for_starter_plan" {
```
glommer, line 42: let's move away from printing "starter" for users, as we now call it free.

internal/turso/utils.go
```diff
@@ -24,10 +24,38 @@ func marshal(data interface{}) (io.Reader, error) {
...
+		return fmt.Errorf("%s are not available on the starter plan - upgrade to access this feature", feature)
```
glommer, line 43:
```suggestion
		return fmt.Errorf("%s are not available on the free plan - upgrade to access this feature", feature)
```

internal/turso/utils.go
notrab: Committed your other change but since this is matching the API response, we'll need to update the API (and Dashboard) to reflect those changes too. I'll create a new issue.

PR #973, opened 2025-04-28
Add `turso db export` command

penberg:
This adds a `turso db export` command, which exports a Turso database to
a local SQLite file. The export command does not generate a WAL file so
the database is a snapshot of time at the beginning of the current
generation, not the latest database version. You need to, therefore,
sync the database using Turso SDK if you need the latest data.

Example usage:

1. Export the `hello` database as `hello.db`:

```
turso db export hello
```

2. Overwrite an existing database file:

```
turso db export hello --overwrite
```

3. Generate metadata with the export:

```
turso db export hello --with-metadata
```

4. Output `hello` database to the `world.db` file:

```
turso db export hello --output-file world.db
```

### DIFF
--- a/internal/cmd/db_export.go
+++ b/internal/cmd/db_export.go
@@ -1,0 +1,66 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var withMetadata bool
+var overwriteExport bool
+var outputFile string
+
+var exportCmd = &cobra.Command{
+	Use:   "export <database>",
+	Short: "Export a database snapshot from Turso to SQLite file.",
+	Long: `Export a database snapshot from Turso to SQLite file.
+
+This command exports a snapshot of the current generation of a Turso database
+to a local SQLite file.  Note that the exported file may not contain the
+latest changes. Use SDK to sync the database after exporting  to ensure you
+have the most recent version.`,
+	Args: cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


internal/cmd/db_export.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+var withMetadata bool
+var overwriteExport bool
+
+var exportCmd = &cobra.Command{
+	Use:   "export <database>",
+	Short: "Export a SQLite database file from Turso.",
```
avinassh, line 16: should we hint that its a snapshot

avinassh (approved):

internal/cmd/db_export.go
penberg:
```$ turso db export --help
Export a database snapshot from Turso to SQLite file.

This command exports a snapshot of the current generation of a Turso database
to a local SQLite file.  Note that the exported file may not contain the
latest changes. Use SDK to sync the database after exporting  to ensure you
have the most recent version.

Usage:
  turso db export <database> [flags]

Flags:
  -h, --help                 help for export
      --output-file string   Specify the output file name (default: <database>.db)
      --overwrite            Overwrite output file if it exists.
      --with-metadata        Include metadata in the export.

Global Flags:
  -c, --config-path string   Path to the directory with config file```

**Comment by npearson72** [comment]
I'm on `turso version v1.0.9` and getting the following error with the export command:

```
Error: failed to export database: response failed with status 404 Not Found
```

Is there another way to get a snapshot/dumb from Turso that I can then import?

PR #987, opened 2025-07-05
internal/cmd/root.go: Handle error when adding reset config flag

12ya:
Refactored AddResetConfigFlag to return an error and updated its usage in root.go to handle potential errors when marking the flag as hidden.

internal/flags/config: refactor accordingly

### DIFF
--- a/internal/flags/config.go
+++ b/internal/flags/config.go
@@ -6,9 +6,12 @@ import (
 
 var resetConfig bool
 
-func AddResetConfigFlag(cmd *cobra.Command) {
+func AddResetConfigFlag(cmd *cobra.Command) error {
 	cmd.PersistentFlags().BoolVar(&resetConfig, "reset-config", false, "")
-	cmd.PersistentFlags().MarkHidden("reset-config")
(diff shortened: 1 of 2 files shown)


internal/flags/config.go
```diff
@@ -6,9 +6,12 @@ import (
 
 var resetConfig bool
 
-func AddResetConfigFlag(cmd *cobra.Command) {
+func AddResetConfigFlag(cmd *cobra.Command) error {
 	cmd.PersistentFlags().BoolVar(&resetConfig, "reset-config", false, "")
-	cmd.PersistentFlags().MarkHidden("reset-config")
+	if err := cmd.PersistentFlags().MarkHidden("reset-config"); err != nil {
```
PThorpe92, line 11: Not sure why this would error? we are adding the flag directly above it.

```go
// from spf13/pflag/flag.go
func (f *FlagSet) MarkHidden(name string) error {
	flag := f.Lookup(name)
	if flag == nil {
		return fmt.Errorf("flag %q does not exist", name)
	}
	flag.Hidden = true
	return nil
}
```

PR #998, opened 2025-12-04
Make it possible to reset baseUrl

LeMikaelF:
This PR makes it possible to reset the CLI's base url by setting `TURSO_API_BASEURL`. I think this was the original intent, but it wasn't working, because the existing config file was overriding the empty environment variable.

### DIFF
--- a/internal/cmd/utils.go
+++ b/internal/cmd/utils.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 
 	"github.com/olekukonko/tablewriter"
 	"github.com/tursodatabase/turso-cli/internal"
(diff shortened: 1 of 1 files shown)


LeMikaelF:
> Hey @LeMikaelF! I didn't understand the scenario where you need this. Why do you need to set `TURSO_API_BASEURL` to empty string?

It's because the base url is persisted in `settings.json`. I had a settings.json that contained my local URL and I wanted to go back to the default one, but I couldn't. I had to figure out where the file is kept (it's pretty deep) and delete it manually.
I knew I had seen the resetting behaviour documented somewhere, it's in api/README.md in turso-platform:

```
# Against your local API server
$ export TURSO_API_BASEURL="http://127.0.0.1:8080/"

# Or against deployed version
$ export TURSO_API_BASEURL=""
```

PR #1012, opened 2025-12-26
Various upload improvements

LeMikaelF:
## Summary

This PR contains various improvements to the multipart upload flow. I realize that it's a lot in the same PR, sorry about that.

* always use multipart upload
* retry chunks when they fail. I expect that this will increase stability compared to the previous flow.
* use the new endpoints, since there was a breaking change (there's an upload_id that's now exposed)
* refresh token between each chunk and before `/finalize`. This was necessary because the new flow can include several requests over more than an hour.
* Show a spinner with the DB size during validation. The first thing we do when we start an upload is that we check that the DB is valid. For large uploads, this could make it look like the CLI is hanging.
* Add debug logging for multipart upload, using the `TURSO_DEBUG_UPLOAD=1` environment variable. I wasn't able to use the existing `--debug` flag, because that one logs the bodies of the PUT requests, and they can contain control characters that will wreck your terminal.
* Update the progress tracker at least every 2 seconds. Previously, it was updated every 1% of an upload. For very large uploads, this made it look like the CLI was hanging.

## Testing

I uploaded a very small databases (few KB), one that was ~100MB, and one that was 1GB. There are also some automated tests in the PR.

## Failing CI test

The integration tests are failing in the CI. It's because multipart upload isn't deployed to us-east-1 yet. I changed the integration tests to use `eu-north-1`. You can see a passing run in that PR (https://github.com/tursodatabase/tursotest/pull/93).

## AI use

The code was mostly written by Claude, but I iterated a lot with it.

### DIFF
--- a/internal/turso/tursoServer.go
+++ b/internal/turso/tursoServer.go
@@ -104,9 +104,6 @@ func (i *TursoServerClient) refreshTokenIfNeeded() error {
 	if i.tokenProvider == nil {
 		return nil
 	}
-	if time.Since(i.lastTokenRefresh) < i.tokenTTL/3 {
-		return nil
-	}
 	token, err := i.tokenProvider()
 	if err != nil {
 		return fmt.Errorf("failed to refresh token: %w", err)


internal/turso/databases.go
```diff
@@ -237,9 +231,21 @@ func (d *DatabasesClient) Create(name, location, image, extensions, group string
...
-	// Create a short-lived DB token for the newly created database to facilitate the upload
-	token, err := d.Token(resp.Database.Name, "1h", false, nil, nil)
+func (d *DatabasesClient) UploadDatabaseAWS(resp *CreateDatabaseResponse, group, uploadFilepath, remoteEncryptionCipher, remoteEncryptionKey string, spinner *prompt.SpinnerT) (*CreateDatabaseResponse, error) {
+	// Get file size to calculate appropriate token expiration
+	fileInfo, err := os.Stat(uploadFilepath)
+	if err != nil {
+		return nil, fmt.Errorf("could not stat upload file: %w", err)
+	}
+
+	// Calculate expiration: assume 10 MB/s minimum upload speed
```
LeMikaelF: This is a security change.

internal/turso/tursoServer.go
```diff
@@ -34,83 +55,221 @@ func (pr *progressReader) Read(p []byte) (int, error) {
...
-// UploadFileSinglePart uploads a database file to the Turso server using a single request.
-// it assumes a SQLite file exists at 'filepath'.
-// it streams the file to the server, and calls the onProgress callback with the progress of the upload.
-func (i *TursoServerClient) UploadFileSinglePart(filepath, remoteEncryptionCipher, remoteEncryptionKey string, onUploadProgress func(progressPct int, uploadedBytes int64, totalBytes int64, elapsedTime time.Duration, done bool)) error {
-	file, err := os.Open(filepath)
+func (i *TursoServerClient) refreshTokenIfNeeded() error {
+	if i.tokenProvider == nil {
+		return nil
+	}
+	if time.Since(i.lastTokenRefresh) < i.tokenTTL/3 {
```
sivukhin: Is it too noticeable overhead if we have 5 minute TTL for the token? I mean that with poor network connection we can refresh token after every chunk - which can create significant overhead.

internal/cmd/group_flag.go
```diff
@@ -307,9 +321,11 @@ func sqliteFileIntegrityChecks(file string, cipher string) error {
 	if flags.Debug() {
 		log.Printf("Running integrity check...")
 	}
-	_, err = exec.Command("sqlite3", file, "pragma quick_check;").CombinedOutput()
+	spinner := prompt.Spinner(fmt.Sprintf("Validating database file (%s)...", humanReadableSize(fileInfo.Size())))
```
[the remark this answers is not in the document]
LeMikaelF: Created an issue for this https://github.com/tursodatabase/turso-cli/issues/1020

PR #1050, opened 2026-06-10
feat: database access allow rules by IP/CIDR and AWS VPC endpoint ID

glommer:
## Summary

Adds CLI support for the database access restrictions recently added to the platform (allow rules by IP/CIDR and AWS VPC endpoint ID), with the same developer experience as database-level delete protection.

New command group:

```
turso db config allow-rules show <db>
turso db config allow-rules set <db> --ip 203.0.113.7 --ip 10.0.0.0/8 --aws-vpc vpce-0fe6c8807461bba49
turso db config allow-rules clear <db> [--ips] [--aws-vpcs]
```

- `set` replaces each list whose flag is provided; a list whose flag is not given is left unchanged. Entries are trimmed, deduplicated, and validated client-side to match the server rules (individual IPs or CIDR blocks via `net.ParseIP`/`net.ParseCIDR`, VPC endpoint ids must start with `vpce-`).
- `clear` clears both lists by default, or selectively with `--ips` / `--aws-vpcs`.
- When both lists are configured, connections must satisfy both (AND semantics): an allowed IP arriving through an allowed AWS VPC endpoint. Help texts spell this out.
- `turso db show` now displays `Allowed IPs` and `Allowed VPC IDs` when configured.

This uses the existing database configuration endpoint (`GET`/`PATCH .../databases/{db}/configuration`) with the `allowed_ips` and `allowed_aws_vpc_ids` fields. The new `DatabaseConfig` fields are `*[]string` with `omitempty`, which is load-bearing: the server treats an explicit `null` as "clear the list", so the fields are only serialized when deliberately set — keeping unrelated config updates (e.g. delete protection) from wiping the allow rules.

Allow rules are database-level only (no group-level support), matching the platform API.

## Test plan

- `go build ./...` and `go vet` pass
- New unit tests for entry normalization and IP/CIDR/VPC-id validation (`go test ./internal/cmd -run 'TestNormalizeAllowRuleEntries|TestValidateAllowed'`)
- Help output for all subcommands verified manually
- Note: `TestGetReservedBytes` fails in `internal/cmd`, but fails identically on a clean checkout of main — pre-existing and unrelated
- Worth a live round-trip against an AWS database before relying on it (server rejects allow rules on Fly databases and the starter plan; the CLI surfaces those errors as-is)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/cmd/db_allow_list.go
+++ b/internal/cmd/db_allow_list.go
@@ -1,0 +1,215 @@
+package cmd
+
+import (
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/tursodatabase/turso-cli/internal"
+	"github.com/tursodatabase/turso-cli/internal/turso"
+)
+
+var allowListIPsFlag []string
+var allowListVpcsFlag []string
+var clearAllowListIPsFlag bool
+var clearAllowListVpcsFlag bool
+
+func init() {
+	dbConfigCmd.AddCommand(dbAllowListCmd)
+	dbAllowListCmd.AddCommand(dbShowAllowListCmd)
+	dbAllowListCmd.AddCommand(dbSetAllowListCmd)
+	dbAllowListCmd.AddCommand(dbClearAllowListCmd)
+	dbSetAllowListCmd.Flags().StringSliceVar(&allowListIPsFlag, "ip", nil, "IP address or CIDR block to allow. Can be repeated. Replaces the current list of allowed IPs.")
+	dbSetAllowListCmd.Flags().StringSliceVar(&allowListVpcsFlag, "vpc", nil, "AWS VPC endpoint ID (vpce-...) to allow. Can be repeated. Replaces the current list of allowed VPC endpoints.")
+	dbClearAllowListCmd.Flags().BoolVar(&clearAllowListIPsFlag, "ips", false, "Clear only the list of allowed IPs")
+	dbClearAllowListCmd.Flags().BoolVar(&clearAllowListVpcsFlag, "vpcs", false, "Clear only the list of allowed AWS VPC endpoint IDs")
+}
+
+var dbAllowListCmd = &cobra.Command{
+	Use:               "allow-list",
+	Short:             "Manage the access allow list of a database",
+	Long:              "Manage the access allow list of a database. When an allow list is configured, only connections from the listed IPs/CIDRs or AWS VPC endpoints are accepted.",
+	ValidArgsFunction: noSpaceArg,
+}
+
+var dbShowAllowListCmd = &cobra.Command{
+	Use:               "show <database-name>",
+	Short:             "Shows the access allow list of a database",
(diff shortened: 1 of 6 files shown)


internal/cmd/db_allow_list.go
```diff
@@ -0,0 +1,215 @@
...
+var clearAllowListIPsFlag bool
+var clearAllowListVpcsFlag bool
+
+func init() {
+	dbConfigCmd.AddCommand(dbAllowListCmd)
+	dbAllowListCmd.AddCommand(dbShowAllowListCmd)
+	dbAllowListCmd.AddCommand(dbSetAllowListCmd)
+	dbAllowListCmd.AddCommand(dbClearAllowListCmd)
+	dbSetAllowListCmd.Flags().StringSliceVar(&allowListIPsFlag, "ip", nil, "IP address or CIDR block to allow. Can be repeated. Replaces the current list of allowed IPs.")
+	dbSetAllowListCmd.Flags().StringSliceVar(&allowListVpcsFlag, "vpc", nil, "AWS VPC endpoint ID (vpce-...) to allow. Can be repeated. Replaces the current list of allowed VPC endpoints.")
```
sivukhin: aws-vpc?

internal/cmd/db_allow_list.go
```diff
@@ -0,0 +1,215 @@
...
+	dbClearAllowListCmd.Flags().BoolVar(&clearAllowListIPsFlag, "ips", false, "Clear only the list of allowed IPs")
+	dbClearAllowListCmd.Flags().BoolVar(&clearAllowListVpcsFlag, "vpcs", false, "Clear only the list of allowed AWS VPC endpoint IDs")
+}
+
+var dbAllowListCmd = &cobra.Command{
+	Use:               "allow-list",
```
sivukhin: maybe allow-rules? at least that's the name I used in the api :)

internal/cmd/db_allow_list.go
```diff
@@ -0,0 +1,215 @@
...
+	Short:             "Manage the access allow list of a database",
+	Long:              "Manage the access allow list of a database. When an allow list is configured, only connections from the listed IPs/CIDRs or AWS VPC endpoints are accepted.",
```
sivukhin: we have AND semantic BTW - wording here can be confusing

sivukhin (approved):

internal/cmd/db_allow_list.go
glommer: Renamed to `--aws-vpc` (and `clear`'s selective flag to `--aws-vpcs`) in 8635573.

internal/cmd/db_allow_list.go
glommer: Renamed the command group to `allow-rules` to match the API in 8635573.

internal/cmd/db_allow_list.go
glommer: Good catch — reworded the help texts to spell out the AND semantics in 8635573.

Issue #318, opened 2023-04-06
Native Windows support


penberg:
We recommend people to use WSL, but there the CLI does work on Windows (with some bugs). If there's sufficient demand, let's fix up the issues to make native Windows a supported platform.

55lj:
FWIW, turso-cli runs perfectly fine on my WSL2 (e.g., I don't face missing urls from https://github.com/chiselstrike/turso-cli/issues/259), I installed via curl and are on `Linux XXX 5.15.90.1-microsoft-standard-WSL2`

Looking at VS Code and all those remote extensions from Microsoft supporting developing in WSL2, I assume that developing in WSL2 on Windows is kinda of the default nowadays.

TigersWay:
> developing in WSL2 on Windows is kind of the default nowadays (unless you do native Windows software).

It may be the default for some, but I - I know for certain I am not alone - usually have more than enough with git-bash.
**Install and use WSL only for Turso is kinda "heavy"**!

henrikvilhelmberglund:
Noticed this too when wanting to try out turso. I would prefer not to use WSL since I don't use it for anything else.
![mintty_D9I9ZB3z0Z](https://github.com/chiselstrike/turso-cli/assets/110549389/e14515a5-b7d6-4206-96e2-88416dae357d)

sakgoyal:
Is there an update for this? I really don't want to install so many dependencies just to connect to my DB.

haaawk:
We're currently busy wrapping up other work. We will take this issue into consideration when planning future work. Sorry for the inconvenience.

thatsmycode:
[...]

PR #1055, opened 2026-07-17
feat: support Windows releases

ryuapp:
related #318

I usually use turso cli on Windows via `go install`, and it works well in most cases.
However, I would appreciate it if you could provide binaries, as that would remove the dependency on Go.
For those using tools such as [`mise`](https://mise.jdx.dev/) or [`aqua`](https://aquaproj.github.io/), this change will make installation even easier.

## Limitation

As a limitation, since [libsql](https://github.com/tursodatabase/libsql) does not provide binaries for Windows, it is currently not possible to set up a development environment using the `turso dev` command.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -13,7 +13,7 @@ permissions:
   contents: read
 
 jobs:
-  build:
+  checks:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4.2.2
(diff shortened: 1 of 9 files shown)


ofek:
@sivukhin @PThorpe92 Could one of you perhaps please press the button to allow CI to run?

penberg:
@claude review

ryuapp:
How is it?

PR #1056, opened 2026-07-22
add export log for Turso databases

pedrocarlo:
Just issues an export command to the server for the tursodb log

### DIFF
--- a/internal/cmd/db_export.go
+++ b/internal/cmd/db_export.go
@@ -13,12 +13,14 @@ var outputFile string
 
 var exportCmd = &cobra.Command{
 	Use:   "export <database>",
-	Short: "Export a database snapshot and WAL from Turso to SQLite files.",
-	Long: `Export a database snapshot and WAL from Turso to SQLite files.
+	Short: "Export a database snapshot and its log from Turso to local files.",
+	Long: `Export a database snapshot and its log from Turso to local files.
 
 This command exports a snapshot of the current generation of a Turso database
-to a local SQLite file, along with any WAL (Write-Ahead Log) frames. The WAL
-file will be saved as <database>.db-wal alongside the main database file.`,
+to a local database file. For SQLite-backed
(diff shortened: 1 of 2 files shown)


internal/cmd/db_export.go
```diff
@@ -13,12 +13,14 @@ var outputFile string
...
 	Use:   "export <database>",
-	Short: "Export a database snapshot and WAL from Turso to SQLite files.",
-	Long: `Export a database snapshot and WAL from Turso to SQLite files.
+	Short: "Export a database snapshot and its log from Turso to local files.",
+	Long: `Export a database snapshot and its log from Turso to local files.
 
 This command exports a snapshot of the current generation of a Turso database
-to a local SQLite file, along with any WAL (Write-Ahead Log) frames. The WAL
-file will be saved as <database>.db-wal alongside the main database file.`,
+to a local database file, along with its pending writes. For SQLite-backed
```
sivukhin: pending writes?

internal/turso/tursoServer.go
```diff
@@ -484,9 +485,19 @@ func (i *TursoServerClient) Export(outputFile string, withMetadata bool, remoteE
...
+		if err := i.ExportLog(outputFile, &info, remoteEncryptionKey); err != nil {
+			return fmt.Errorf("failed to export logical log: %w", err)
+		}
+	} else {
+		lastFrameNo, err = i.ExportWAL(outputFile, &info, remoteEncryptionKey)
+		if err != nil {
+			return fmt.Errorf("failed to export WAL: %w", err)
+		}
 	}
 	if withMetadata {
```
sivukhin: probably we should do this only for non tursodb for now

sivukhin (approved):

internal/cmd/db_export.go
pedrocarlo: BS let me change sorry

internal/turso/tursoServer.go
pedrocarlo: ok

PR #1062, opened 2026-07-27
Support TursoDB database imports

pedrocarlo:
## Summary

- detect rollback, WAL, and MVCC SQLite file formats before import
- safely checkpoint sidecars and convert compatible uploads to TursoDB format
- default `db import` to TursoDB and route file seeds through the upload flow
- add coverage for file-format detection, sidecar handling, and multipart conversion

## Why

TursoDB imports could be routed through an incompatible create path or rejected because the uploaded file format did not match the target database. The import flow now validates and converts supported inputs without silently dropping sidecar data.

## Validation

- `go test ./internal/turso` passes
- `go test ./internal/cmd` currently fails in pre-existing `TestGetReservedBytes`: expected 0 reserved bytes, got 12


internal/cmd/db_create.go
```diff
@@ -155,7 +155,12 @@ func createDatabase(client *turso.Client, name, location, groupName string, seed
 	if sizeLimitFlag != "" {
 		return createDatabaseV2(client, name, location, groupName, seed, spinner)
 	}
-	if seed != nil && seed.Type != "database" && seed.Type != "upload" {
+	// Only fork seeds ("database") can go through the v3 API. File seeds
+	// ("database_upload") must use the v2 flow: the v3 branch never uploads
+	// the file after creating the database. (This used to compare against
+	// "upload", a seed type that doesn't exist, which routed uploads to v2
+	// by accident.)
```
LeMikaelF:
```suggestion
```

internal/cmd/group_flag.go
```diff
@@ -340,11 +341,158 @@ func runQuickCheck(file string) error {
 	return nil
 }
 
+// checkpointWALBeforeUpload folds any pending WAL frames into the main
+// database file and verifies no data is left behind in sidecar files. The
+// upload ships only the main database file, so frames still sitting in
+// <file>-wal (or a hot rollback journal) would silently be missing from the
+// imported database.
```
LeMikaelF:
```suggestion
```

internal/cmd/group_flag.go
```diff
@@ -340,11 +341,158 @@ func runQuickCheck(file string) error {
...
 		return nil, err
 	}
 
+	if format == fileFormatRollback {
+		if err := checkSQLiteAvailable(); err != nil {
+			return nil, err
+		}
+		// The server only accepts WAL or MVCC format files. Converting to WAL
+		// is exactly the remediation the error message used to instruct users
+		// to run themselves, and sqlite3 is already a hard requirement here.
```
LeMikaelF:
```suggestion
```

internal/cmd/group_flag.go
```diff
@@ -340,11 +341,158 @@ func runQuickCheck(file string) error {
...
+			return err
+		}
+	}
+	return nil
+}
+
+// prepareTursoDBFile asks the TursoDB engine to convert the database to MVCC,
+// checkpoint any logical-log entries into the main file, and validate the
+// resulting database. Uploading only the main file is safe once all data-bearing
+// sidecars are empty.
```
LeMikaelF:
```suggestion
// resulting database. Uploading only the main file is safe once the logical log
// and WAL have been truncated.
```

internal/cmd/db_import.go
```diff
@@ -9,11 +9,16 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// Import keeps its own flag value because it delegates database creation to
+// CreateDatabase; RunE copies the selected value to the shared create flag.
```
LeMikaelF:
```suggestion
```

internal/cmd/group_flag.go
```diff
@@ -340,11 +341,158 @@ func runQuickCheck(file string) error {
...
+func prepareTursoDBFile(file string) error {
+	output, err := exec.Command("tursodb", "-q", "-m", "list", file,
+		"PRAGMA journal_mode = mvcc; PRAGMA wal_checkpoint(TRUNCATE); PRAGMA quick_check;").CombinedOutput()
```
LeMikaelF: In https://github.com/tursodatabase/turso-cli/pull/1012, I added a spinner during `quick_check`, because for large DBs, it can make appear like the CLI is hanging. With this change, it doesn't look like there's a spinner.

See the `runQuickCheck()` function.

internal/cmd/group_flag.go
```diff
@@ -340,11 +341,158 @@ func runQuickCheck(file string) error {
...
+		}
+	}
+	return nil
+}
+
+// prepareTursoDBFile asks the TursoDB engine to convert the database to MVCC,
+// checkpoint any logical-log entries into the main file, and validate the
+// resulting database. Uploading only the main file is safe once all data-bearing
+// sidecars are empty.
+func prepareTursoDBFile(file string) error {
```
LeMikaelF: it's weird that the checkpointing/preparing/pre-upload logic is in a file called `group_flag.go` 🤔

internal/cmd/group_flag.go
```diff
@@ -340,11 +341,158 @@ func runQuickCheck(file string) error {
...
+	}
+	if info.Size() > 0 {
+		return fmt.Errorf("%s is not empty, importing would lose the data it holds: %s", sidecarPath, hint)
+	}
+	return nil
+}
+
+// tursodbLogPath returns the logical-log sidecar path tursodb uses for a
+// database file, mirroring turso_core's `with_extension("db-log")`: the
+// file's extension (if any) is replaced with "db-log".
```
LeMikaelF:
```suggestion
```

internal/cmd/group_flag.go
```diff
@@ -340,11 +341,158 @@ func runQuickCheck(file string) error {
...
+	case fileFormatMVCC:
+		// sqlite3-based checks can't run on MVCC (tursodb format) files: the
+		// sqlite3 binary reports them as not-a-database.
+		if !tursoDBFlag {
+			return nil, fmt.Errorf("%s is in tursodb (MVCC) format and can only be imported into a tursodb database", file)
+		}
+		if cipher != "" {
+			return nil, errors.New("remote encryption is not supported when importing tursodb (MVCC) format files")
+		}
+	case fileFormatNotSQLite:
```
LeMikaelF: `fileFormatUnknown` is not handled

internal/cmd/group_flag.go
```diff
@@ -340,11 +341,158 @@ func runQuickCheck(file string) error {
...
+
+	switch format {
+	case fileFormatWAL:
+		if err := checkSQLiteAvailable(); err != nil {
+			return nil, err
+		}
+		if err := checkpointWALBeforeUpload(file); err != nil {
+			return nil, err
+		}
+		if err := sqliteFileIntegrityChecks(file, cipher); err != nil {
```
LeMikaelF: This function contains some nice code (a good error message if the user uploads a dump, spinner on quick_check, max size check, page size, etc.) that isn't replicated in the Turso code. Maybe reuse this function with turso instead of sqlite?

internal/cmd/group_flag.go
```diff
@@ -340,11 +341,158 @@ func runQuickCheck(file string) error {
...
+		}
+		if format != fileFormatMVCC {
+			fmt.Printf("Converting %s to TursoDB (MVCC) format for import.\n", file)
+		}
+		if err := prepareTursoDBFile(file); err != nil {
+			return nil, err
+		}
+	}
+
+	fileInfo, err := os.Stat(file)
```
LeMikaelF: This check is now performed twice for WAL-mode databases (other one is in `sqliteFileIntegrityChecks`).

internal/cmd/group_flag.go
```diff
@@ -397,6 +555,14 @@ func checkSQLiteAvailable() error {
 	return err
 }
 
+func checkTursoDBAvailable() error {
+	_, err := exec.LookPath("tursodb")
+	if errors.Is(err, exec.ErrNotFound) {
+		return errors.New("could not find tursodb on your system. Please install it to import into a TursoDB database")
```
LeMikaelF: This could be a sub-par first experience for a new user looking to import their DB into the cloud, because they'll need to install `turso`, then get an error, and figure out how to download `tursodb`.

Could we give them instructions on installing tursodb?

I wonder if there's a way we could avoid this. It doesn't make sense for a user to have to install 2 CLI's.

pedrocarlo: we could use `turso-go`. I did this because we also require the user to have sqlite3 installed. Don't know if this introduces any issues for building the cli.

internal/cmd/db_import_file.go
```diff
@@ -0,0 +1,446 @@
...
+	_ "turso.tech/database/tursogo"
+)
+
+const MaxAWSDBSizeBytes = 1024 * 1024 * 1024 * 20 // 20 GB
+
+const databaseSettingsQuery = "select journal_mode as j, page_size as p, auto_vacuum as a, encoding as e from pragma_journal_mode, pragma_page_size, pragma_auto_vacuum, pragma_encoding;"
+
+type databaseFileChecker struct {
+	name        string
+	binary      string
```
LeMikaelF: nit: this type is now weird, because the `binary` field either has value `sqlite3` or `""`

LeMikaelF (approved):

internal/cmd/db_import_file.go
pedrocarlo: yeah i've seen it and will adjust it, Im first just trying to the run code locally first to confirm it works, then will move to an interface of functions probably

pedrocarlo: or maybe just an int to identify

pedrocarlo: ok changed it now

